Issue #4, opened 2012-03-29
stale versions when running with setuptools


warner:
Zooko and I noticed (at Pycon, while versioneer-ifying pycryptopp) that the following sequence:

* setup.py build
* make some changes
* setup.py install

resulted in installed code that had the right version string, but the copy in `build/` was stale (I vaguely remember the code in build/ being updated but the version string remaining unmodified). I don't know exactly what was happening, but I remember it making me question my previous assumption that `setup.py install` would always execute `setup.py build` (and the associated versioneer hook) first.

warner:
This hit me again today when using `setup.py install` or `develop` on a top-level package (in this case, `magic-wormhole`) which depended upon a versioneer-enabled second package (`spake2`). With versioneer-0.12, the second package got a version of "unknown". With the current versioneer (0.14+, b28e729) it's worse: it gets the same version as the top-level package, which is really really broken.

It turns out that dependencies are handled by running `setup.py bdist_egg` on their unpacked tarballs, in a python process where the imports table has been hacked up to replace distutils with setuptools. So, as noted before, either the upcalls are going to the wrong place, or our overridden functions aren't getting called at all.

warner:
I think this should work now: a combination of deriving sdist from the right place (df3eb3e) and overriding `build_py` instead of `build` (28e8376). The new 0.15 release includes both. Please re-open if you see any problems.

PR #8, opened 2013-02-28
Hook build_py instead of build

jimparis:
This seems to help issue #4


warner:
This has been applied (in a different form) in 28e8376. `build_py` was indeed the right thing to hook, but it took me a long time to understand the internal setuptools flow to believe it :). I'm still not sure about what happens with script-only or C-only packages, but I guess we'll fix those when someone has a use for them and files a bug.

thanks!

PR #9, opened 2013-08-12
PEP-0440 compatibility: add git-describe's commit-distance as post-release/dev-release info

muggenhor:
PEP-0440 compatibility: add git-describe's commit-distance as post-release/dev-release info

This automatically creates post-release, development releases for the tag found *and* ensures they're PEP-0440 compatible to ensure tools like `pip` can work with them.


warner:
I think PR #67 took care of this.

PR #45, opened 2014-07-30
added pep440 compliance option

sebastianneubauer:
Hi, we have added some functionalty to replace the version string git commit number with a more pep440 compliant 'dev'. There is also the option to replace 'dev' with a string of your choice (e.g. 'post'). This can be configured via the two new header.py variables:

    pep440_compliant (default False)
    release_type_string (default 'dev')

We can see there is some major refactoring in progress in the 'template' branch, but we hoped it would be possible to incorporate these small updates to the master. No test for 'git2pep440', but can be added if you agree to incorporate the changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,8 +6,8 @@ The Versioneer
 * Brian Warner
 * License: Public Domain
 * Compatible With: python2.6, 2.7, 3.2, 3.3, 3.4, and pypy
-
-[![Build Status](https://travis-ci.org/warner/python-versioneer.png?branch=master)](https://travis-ci.org/warner/python-versioneer)
+* [![Latest Version](https://pypip.in/version/versioneer/badge.svg)](https://pypi.python.org/pypi/versioneer/)
+* [![Build Status](https://travis-ci.org/warner/python-versioneer.png?branch=master)](https://travis-ci.org/warner/python-versioneer)
 
 This is a tool for managing a recorded version number in distutils-based
 python projects. The goal is to remove the tedious and error-prone "update
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ The Versioneer
 * Brian Warner
 * License: Public Domain
 * Compatible With: python2.6, 2.7, 3.2, 3.3, 3.4, and pypy
-* [![Latest Version](https://pypip.in/version/versioneer/badge.svg)](https://pypi.python.org/pypi/versioneer/)
+* [![Latest Version](https://pypip.in/version/versioneer/badge.svg?style=flat)](https://pypi.python.org/pypi/versioneer/)
 * [![Build Status](https://travis-ci.org/warner/python-versioneer.png?branch=master)](https://travis-ci.org/warner/python-versioneer)
 
 This is a tool for managing a recorded version number in distutils-based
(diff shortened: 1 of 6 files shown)


matthew-brett:
Is there anything I / we can do to help move this along? This would be very useful for me.

src/get_versions.py
```diff
@@ -36,27 +36,52 @@ def get_versions(default=DEFAULT, verbose=False):
...
     if verbose: print("got version from default %s" % default)
     return default
 
 def get_version(verbose=False):
     return get_versions(verbose=verbose)["version"]
+
+
+def git2pep440(ver_str):
+    try:
+        tag, commits, _ = ver_str.split('-', 2)
```
pitrou: +1 on @tacaswell's proposal. I don't think there's any need to make it optional, not more than in the current versioneer behaviour.

glennmatthews:
As far as I can tell this pull does not propagate `pep440_compliant` into `_version.py`, so `_version.py` `get_versions()` still reports the non-PEP440 version.

sebastianneubauer:
I found that out recently too and started to fix it, but then found out, that versioneer is far more complicated than I expected at first glance. I would need some more guidance on how to fix it...BTW is the project still active, i.e is it worth the time to investigate it? Is this feature still needed for someone?

matthew-brett:
I certainly do have use for this feature, and I have otherwise found versioneer very useful.  I hope the project does keep going.

glennmatthews:
Getting _version.py to also do the right thing seems to just require making a similar set of updates to `src/git/long_get_versions.py`. I don't like the duplication of code that's involved in my own version (glennmatthews@711bd3eb) but it does at least work.

sebastianneubauer:
Am I wrong, or is the part behind '+' just ignored by pip? Would that mean that all versions after a tagged version are equal? How can I prevent being installed by pip only with --pre if not a tagged version? The exact string between version and commit distance is configurable, default is '.post.dev', changing it to '.post' would make the version a stable one if you need this behaviour, a,b, rc wold be other options...BTW: this revision works now for me and is now the default, pep440 can be switched of via configuration

tacaswell:
I think the part after the `+` is sorted, but there are no rules or validation on it's format (the idea is to let packagers keep track of their internal patch levels (ex 2.3.1+ubuntu3)

warner:
Ok, I think PR #67 subsumed this. The new template branch should make it possible to get other formats, including .post.dev, so once that lands, I think it should be easy to get the workflows above.

README.md
```diff
@@ -6,8 +6,8 @@ The Versioneer
 * Brian Warner
 * License: Public Domain
 * Compatible With: python2.6, 2.7, 3.2, 3.3, 3.4, and pypy
-
-[![Build Status](https://travis-ci.org/warner/python-versioneer.png?branch=master)](https://travis-ci.org/warner/python-versioneer)
+* [![Latest Version](https://pypip.in/version/versioneer/badge.svg?style=flat)](https://pypi.python.org/pypi/versioneer/)
```
ac38618635-hue, line 9: ac

ac38618635-hue (approved):

PR #53, opened 2014-11-14
FIX: Use / instead of os.sep

FlorianWilhelm:
In `_version.get_versions()` the root directory is determined by splitting `versionfile_source`. According to the documentation this is something like `package/_version.py`, i.e. separated by slashes. Since `os.sep` is used in `get_versions()` this string will be splitted by a backslash which is wrong. Since most Python packages should work on Windows and Linux, the versioneer configuration should follow the slash convention to separate directories and files.


warner:
Good catch. Landed (in rebased form) in fa3e11ec

Issue #52, opened 2014-11-12
using versioneer in multiple setuptools dependant projects leads to interchanged version numbers


marscher:
I know only distutils is supported. But still I wanted to report this (maybe to avoid others run into this). I will try to figure out whats going on.

Lets say A depends B (both using versioneer). B is in on pypi with a source tarball, so have a static _version.py file.

A drags in B during setup phase, and version of B gets replaced by version of A.
Can be observed here: https://travis-ci.org/markovmodel/PyEMMA/builds/40644627#L623 

Maybe this also occurs if A and B are using distutils (have not tested).

warner:
Oh, that's tricky.. so the `setup.py` from both projects is being evaluated in the *same* python process? And the setup.py's `import versioneer` is picking up the wrong version of versioneer? And by "wrong", I guess we mean "whichever `versioneer.py` got imported first is used for all other builds".

Hm. I see the problem, and it's ugly, but I'm hesitant to dive into force-reloads and module-loader internals. Is it possible to have `setup.py` say `from . import versioneer`, or is that illegal when we're running it as `python setup.py`? Hrm, I wish there were a cleaner solution..

marscher:
If versioneer is on same level as setup.py it should work as expected.

warner:
Hm. Maybe we could do:

* setup.py defines a dict of config data, and *doesn't* set `versioneer.stuff`
* `cmdclass = versioneer.get_commands(configdict)`

Then the config info could get curried into the command classes, and there wouldn't be any global state to get confused between separate uses of the same module.

warner:
If we do it that way, we'd also need `version=versioneer.get_version(configdict)` in the args to `setup()`.

What if we put a `.versioneer.ini` file in the top directory? The setup.py changes would then be just:

```python
import versioneer
cmdclass=versioneer.cmdclass()
setup(... version=versioneer.get_version(), cmdclass=cmdclass, ...)
```

We can probably make that backwards-compatible too, by requiring those values to be set if a .ini file isn't found. Or add a `[versioneer]` section to an existing `setup.cfg` file, instead of using a new file.

marscher:
I vote for the [versioneer] section in setup.cfg. This seems to be the right solution!

warner:
FYI, #74 is where this is happening. Give it a whirl and see if it does what you need. (might still be a bit touchy, let me know if you run into problems)

warner:
[...]

PR #54, opened 2014-11-20
[Readme] changed import statement to support multiple dependant setuptools projects.

marscher:
Fix for #52. I've tested this with 2 setuptools projects A, B, where A depends B and both use versioneer.


warner:
I just landed the change (1fe079a) to use setup.cfg instead of the funky set-attributes-on-imported-versioneer, so I think we no longer need this one. Please let me know if the new scheme doesn't enable the multiple-projects use case you made this for and we'll figure out how to accomodate it.

PR #57, opened 2014-12-15
Cope with symlinks.

rhattersley:
Allows versions to continue working for a checked out source tree when the package was discovered by Python via a symlink.

### DIFF
--- a/src/git/long_get_versions.py
+++ b/src/git/long_get_versions.py
@@ -11,7 +11,7 @@ def get_versions(default={"version": "unknown", "full": ""}, verbose=False):
         return ver
 
     try:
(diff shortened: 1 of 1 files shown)


warner:
Sounds reasonable.. can you think of a way to add a test for this? Is something like `git checkout PROJECT; ln -s PROJECT foo; cd foo; python setup.py version` enough? Or does the symlink need to live below the source-tree directory to cause problems?

I was thinking maybe we could add something to `test_git.py`, to see if symlinks are available, and then link to a source checkout and run an extra copy of the tests.

warner:
Landed (in rebased form) in 66c9f6a. If we can come up with a test, let's add that in a different PR.

PR #59, opened 2014-12-18
Add support for vcs root directory being a parent of module build root.

mklein0:


mklein0:
I have an alternate implementation of this fix which relies on implicitly deriving the GIT root via the **git rev-parse --show-toplevel** command.  This is located on an alternate branch, mklein0/python-versioneer/tree/vcs_root_git_rev_parse@4bc41f84f1fe0934e6bf385dc9fc2a6140386928

warner:
I think this is mostly subsumed by #61, which I've been hacking on the last few days (and which is slightly smaller because I think we don't really need to know exactly where the `.git` directory is: except possibly for installation, Versioneer's git commands can be run from anywhere under the source tree). But I'm adding this to the 0.17 milestone so I don't forget to incorporate the test-coverage and `-w wheelhouse` fixes from this branch.

warner:
Ok, it looks like all the tests from this branch are included in the other one. Closing this in favor of #61.

Issue #38, opened 2014-05-09
Python setup  can not be located in a sub-directory of a repository


mklein0:
The versioneer script does not allow the creation of python modules which are sub-directories of a git repository.

ex:

    repo/
       .git
       module1/
           setup.py
           versioneer.py
               module1/
              __init__.py
               stuff.py
        module2/
           setup.py
           versioneer.py
           module2/
               __init__.py
               morestuff.py

While this directory structure may not be a best practice, it can be a useful construct when building rapid prototypes.

mcbridejc:
I hit the same issue here. Would be a nice improvement.

warner:
Now that 0.15 is out, we should look at fixing this. I'll take a look at #61 and see if I can adapt it to the (large) changes that have landed since then. Thanks!

wulmer:
Would be great if that went into the next release! I love `versioneer` and use it heavily in dozens of projects, but unfortunately there is one with a nested directory structure :-(

isnok:
+1

warner:
Note that current pip-8.1.1 has a bug/issue that causes `pip install .` from a sub-directory-based `setup.py` to fail (it always gets a version of "0+unknown"). `pip install -e .` and `pip wheel` should work, though.

If you see other problems with subprojects like this, please open a new bug.. I wouldn't be surprised if there are other corner cases that are broken somehow. And if current master seems to work for you, please chime in here; it could definitely use more real-world testing than what I've been able to do locally.

PR #61, opened 2015-01-10
Add support for locating setup.py below the git repo root

mcbridejc:
This addresses issue #38. 

versioneer is the best way that I've come across to manage automated versioning of python projects. I know it isn't the standard practice, but I do have a case where I have a python module that is part of a larger project that lives within a single repo, and I wanted to be able to use versioneer for this as well. 

This does a few things: 
1) Adds a test case where the git repo is created in `demoapp`, but the setup.py and other files are created in `demoapp/project`
2) In git_versions_from_vcs(), it uses git command return to test for a git directory, rather than looking for a `.git` directory
3) In versions_from_parentdir(), it will search up two levels of directory hierarchy trying to find a directory name that matches the configured prefix.

### DIFF
--- a/src/from_parentdir.py
+++ b/src/from_parentdir.py
@@ -1,11 +1,18 @@
 
 def versions_from_parentdir(parentdir_prefix, root, verbose=False):
     # Source tarballs conventionally unpack into a directory that includes
-    # both the project name and a version string.
-    dirname = os.path.basename(root)
-    if not dirname.startswith(parentdir_prefix):
-        if verbose:
-            print("guessing rootdir is '%s', but '%s' doesn't start with prefix '%s'" %
-                  (root, dirname, parentdir_prefix))
-        return None
-    return {"version": dirname[len(parentdir_prefix):], "full": ""}
+    # both the project name and a version string. We will also support searching
+    # up two directory levels for an appropriately named parent directory
+    rootdirs = []
+
+    for i in range(3):
+        dirname = os.path.basename(root)
+        if dirname.startswith(parentdir_prefix):
+            return {"version": dirname[len(parentdir_prefix):], "full": ""}
+        else:
+            rootdirs.append(root)
+            root = os.path.dirname(root) # up a level
+    if verbose:
+        print("Tried directories %s but none started with prefix %s" %
+            (str(rootdirs), parentdir_prefix) )
+    return None
--- a/src/git/from_vcs.py
+++ b/src/git/from_vcs.py
@@ -4,15 +4,16 @@ def git_versions_from_vcs(tag_prefix, root, verbose=False):
     # if the git-archive 'subst' keywords were *not* expanded, and
     # _version.py hasn't already been rewritten with a short version string,
     # meaning we're inside a checked out source tree.
-
-    if not os.path.exists(os.path.join(root, ".git")):
-        if verbose:
-            print("no .git in %s" % root)
-        return {}
-
     GITS = ["git"]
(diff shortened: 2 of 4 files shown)


src/from_parentdir.py
```diff
@@ -1,11 +1,18 @@
...
-        if verbose:
-            print("guessing rootdir is '%s', but '%s' doesn't start with prefix '%s'" %
-                  (root, dirname, parentdir_prefix))
-        return None
-    return {"version": dirname[len(parentdir_prefix):], "full": ""}
+    # both the project name and a version string. We will also support searching
+    # up two directory levels for an appropriately named parent directory
+    rootdirs = []
+
+    for i in range(3):
```
wulmer: Is there a special reason for the limitation to 3 parent directories? What if I have a deeper nested structure?

wulmer:
I just tried it with my project setup and it worked for `python setup.py build` being executed in a subdirectory of my project. Great work!

One thing that did not work is the retrieval of the version when using the development version, e.g. when I just run `python` and then import my lib and check its version (`import mylib; print(mylib.__version__)`. This returns `'unknown'` as version.

For me this isn't a blocker, but perhaps there is an easy solution to make this work even in development mode?

warner:
I've rebased this branch and added some fixups (in "61-subprojects"). It's not entirely working for me yet: `python setup.py version` shows the right thing, but a `pip install .` does not (it gets `0+unknown`, which is how versioneer says it can't find a git repo). I haven't tried `pip install --editable .` yet. But I'm working on it.

warner:
So, I now understand why `pip install .` is getting "0+unknown". It turns out that the very first thing `pip install` does is to copy `os.path.dirname(setup_py)` into a temporary directory, and then do everything else (egg_info, bdist_wheel) over there. Unfortunately, when our setup.py is in a subdirectory of the git tree, that copy doesn't include the `.git/` directory like it usually does. So pip winds up doing a `setup.py egg_info` in a `.git`-less copy, and can't find any version data to work with.

`python setup.py install` would work (which uses eggs), as does `pip install --editable .` (since that doesn't try to make a copy first). But not `pip install .`

I don't know how to fix this. @dstufft says they want to make `pip install .` do a `setup.py sdist` instead of a copy, which would do the trick, but I don't know if there's a pip ticket open for that (https://github.com/pypa/pip/issues/2195 is related), or what their timeline might be.

I'll add tests to the 61-subprojects branch to exercise the failing cases.

warner:
https://github.com/pypa/pip/pull/3176 might be the pip PR that could fix this

warner:
I've added the extra tests to 61-subprojects. As expected, `pip install` from a "subproject" tree (where setup.cfg is not in the git root) fails, by getting a "0+unknown" version.

At this point, I think we're either waiting on a fix in pip, or we accept the test failures (mark them as TODO or something) and put a note in the docs that says you can't use "pip install ." from such a project. Unfortunately I think pip-install is the main use case for this, so I don't know if it's worth landing until that's fixed. On the other hand, all forms of install from a subproject tree are broken in versioneer-0.16, so maybe it's a step forward even though it's not exactly useful yet.

warner:
Ok, I've marked the relevant tests as "expected to fail", and I've added a section to the README about the limitations (basically that `pip install .` from a subproject directory will not get a useful version string, until/unless `pip` gets changed). I'm gonna land that branch, since I think it's still useful (especially since `pip install --editable .` ought to work). Please let me know if it seems to work for you too.

PR #63, opened 2015-01-30
Added command to replace _source.py with static version 

ndegroot:
This is automatic using 'build' and 'sdist', is useful for deploying with git.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,11 @@
 /MANIFEST
 /dist/
 /versioneer.egg-info/
+.name
+encodings.xml
+misc.xml
+modules.xml
+python-versioneer.iml
+vcs.xml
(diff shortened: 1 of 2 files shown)


warner:
I pushed a branch named "cmd-static-version" with an updated form of this command (`cmdclass.py` was rearranged a lot since you wrote this patch). Could you give it a spin and see if it meets your needs? I left out the .gitignore changes. I'd also like to add a test before we land it, and maybe rename the command to something more active sounding like "set-static-version" or "write-static-version", or give it a docstring or something, so that `setup.py --help-commands` makes it clear what will happen if you use it.

Callek:
Closing this stale PR. (We can reconsider an update or if there is still a need)

Issue #56, opened 2014-12-02
_version.py should conform to PEP8


roskakori:
Currently versioneer creates a `_version.py` with a few minor PEP8 violations. This is slightly annoying for projects where the build process includes a PEP8 validation. It can easily be fixed using

    $ autopep8 -i _version.py

Nevertheless it could be nice if versioneer builds a clean `_version.py` out of the box.

warner:
Thanks for the fix. I added a .travis.yml change to run `flake8` over the generated `_version.py` so we'll catch any future problems. I wouldn't mind making `versioneer.py` itself PEP8-compatible.. have you run into problems because of complaints therein? Or do you avoid that by not running e.g. flake8 on the top-level files like `setup.py` and `versioneer.py`?

I started to fix `versioneer.py` (which really means fixing the various source files that are used to generate it), and immediately got stuck with the travis "Build Status" image link in the README.md (which gets copied into the `versioneer.py` docstring), as the line is too long. I don't know how to split it without breaking the link. Any ideas?

roskakori:
Thanks for improving PEP8 compatibilty and considering to further improve it.

Your assumption is correct, I'm currently excluding `versioneer.py` from the flake8 check in my project, so it's not a big deal right now.

I'm not sure if I understand the issue with you image link correctly. AFAIK flake8 does not check the README.md so I assume you are referring to Python code generating a REAME.md?

First, you can set the maximum line length for flake8 using a command line option or ini file. I never saw a point to limit my code to 79 chars on a 25 inch monitor, so I use:

```ini
[flake8]
exclude = cutplace/_version.py
ignore = E127,E128,E501
max-line-length = 159
```
Also there are several ways to split a long string across multiple lines in Python, e.g.
```python
x = 'a string ' \
   + 'spawning multiple ' \
   + 'lines.

x = ''.join([
    'a string',
    ' spawning multiple ',
    'lines'
])

# Correct but confusing to many people.
x = 'a string ' \
    'spawning multiple ' \
    'lines'
```
Does this help in solving you issue?

warner:
> I'm not sure if I understand the issue with you image link correctly. AFAIK flake8 does not check the README.md so I assume you are referring to Python code generating a REAME.md?

It's the other way around: `versioneer.py` is constructed by combining several source files, most of which are python, but it also embeds the README.md as a large triple-quoted docstring (look at the beginning of version.py).

[...]

PR #66, opened 2015-02-10
Make _version.py and versioneer.py pass pep8 1.6 (#65)

### LINKED ISSUE #65: versioneer.py should be PEP8-compliant too
(spin-off of #56, which only fixed `_version.py`)

It'd be nice if the generated `versioneer.py` were PEP8-compliant. The biggest obstacle is the overlong lines in the docstring, which are copied from the README, which contain URLs for badges (travis, coveralls) and aren't easily splittable. The fix is probably to just omit those lines when building the docstring.

glennmatthews:
Fix #65 - make `versioneer.py` pass `pep8` 1.6.1, including checking this as part of Travis run.
Also, `pep8` 1.6 adds some new checks such that `_version.py` is now failing again despite #56 being fixed - this makes the necessary changes for `_version.py` to pass `pep8` 1.6 as well.


warner:
I landed the PEP8 ones (rebased slightly). I'm going to try to combine your PEP440 change with the other PRs currently pending. Thanks!

glennmatthews:
Yeah sorry - I didn't mean to include the PEP440 changes in this pull request. That was going to be a subsequent request. :-) Thanks!

PR #72, opened 2015-04-28
prepare for configurable styles

warner:
Not really meant for merging yet


warner:
replacing this with a better sequence

PR #75, opened 2015-05-04
script name versioneer-installer

kristi:
`versioneer-installer` is the script to use, not `versioneer install`


warner:
The script should now be just plain `versioneer`.. that old `versioneer-installer` got removed in 0d1c1bd. We don't have a new release yet, though, so at the moment the README on github/master doesn't match the one on pypi (which still has 0.14).

Did I miss something?

warner:
0.15 has been released, with `versioneer` as the executable script it produces.

PR #78, opened 2015-06-02
Filter git tags by tag_prefix.

### LINKED ISSUE #77: Multiple tags result in "unknown" even if correct tag exists
We want to use versioneer to version multiple python packages in a single git repo.

I have 2 setup.cfgs with:

```
tag_prefix=ExampleA-
```
and
```
tag_prefix=ExampleB-
```

In my case I have 2 directories with versioneer installed. And I tag a single commit twice:

```
git tag ExampleA-1.2.3
git tag ExampleB-2.3.4
```

I was assuming that doing an sdist upload would result in the ExampleA being versioned as 1.2.3 but it goes up as "unknown".

Turning on verbose I get 

```
tag 'ExampleB-2.3.4' doesn't start with prefix 'ExampleA-'
```

Looking at the code it looks like this is from the use of "git describe" which takes the most recent commit. So I suppose this would need to be changed to look at all the tags, not just the most recent one.

Do you envisage supporting this use case? Or am I missing something?

DanLipsitt:
Only look at tags that match `tag_prefix`. Might fix #77.

### DIFF
--- a/src/git/from_vcs.py
+++ b/src/git/from_vcs.py
@@ -21,10 +21,11 @@ def git_pieces_from_vcs(tag_prefix, root, verbose, run_command=run_command):
     GITS = ["git"]
     if sys.platform == "win32":
         GITS = ["git.cmd", "git.exe"]
-    # if there is a tag, this yields TAG-NUM-gHEX[-dirty]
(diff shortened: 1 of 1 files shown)


warner:
looks good.. I'll add a test.

PR #87, opened 2015-06-29
fix problem with pip install not installing non .py files

ldath:
Without this change I had my non .py files in tar.gz but `pip install package.tar.gz` was not coping those non .py files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@
 /MANIFEST
 /dist/
 /versioneer.egg-info/
+/.idea
--- a/src/cmdclass.py
+++ b/src/cmdclass.py
@@ -58,7 +58,11 @@ def get_cmdclass():
     #  setuptools/install -> bdist_egg ->..
     #  setuptools/develop -> ?
 
-    from distutils.command.build_py import build_py as _build_py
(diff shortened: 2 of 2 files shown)


warner:
Good catch. Could you give me a version of the pull-request without the .gitignore change?

ldath:
Why there is a problem with adding `/.idea` automatically generated by PyCharm to `.gitignore`? This prevent adding by accident this folder to the repository.

warner:
Eh, good point. I'll land this.

I've never heard of an `.idea` subdir before: I get a lot of pull requests that include unintended changes, especially to `.gitignore`, where folks add lines for their local tools, and then submit the real code changes along with the extra random stuff. I usually recommend that these lines go into your global `~/.gitignore` instead of putting them into each and every project (e.g. I use emacs, which creates leftover files that end in "~", but I keep the .gitignore line for that in my global ignorefile instead of in the local tree).

Merged (in rebased form) in 99d5d93.

ldath:
In my case was too many situations when a colleague added this folder to the repository disrupting my and others work. That is why adding this folder has come to be my habit. Sorry for the inconvenience and thanks for the explanation.

PR #88, opened 2015-06-29
Versioneer pep257

glennmatthews:
* `versioneer.py` is now also PEP257 compliant
* use [flake8-docstrings](https://pypi.python.org/pypi/flake8-docstrings/0.1.2) to run `pep257` as part of the `flake8` run.
* exclude `.pyc` files from `.gitignore` for convenience.


warner:
Looks good.. merged (in rebased form) in 12e7d8b. Thanks!

PR #94, opened 2015-08-09
added pep440-bare style

clearclaw:
Added a pep440-bare style to address legacy projects which already used the commit count as the unadulterated patch/minor number.


clearclaw:
Gahh!

PR #95, opened 2015-08-09
Added an additional pep440-bare style

clearclaw:
This is useful for legacy projects which already used the count from the last tag as their patch/minor number.  Moving to "post#" in the version string causes sorting problems for them.


clearclaw:
The Travis errors appear to be due to pre-existent pyflake errors.  I can fix those too if wished.

effigies:
We're working on reviving this project. Can you rebase/merge master to resolve conflicts and refresh the CI?

PR #105, opened 2015-11-27
Use only numeric versions in Git, ignore other tags with the same prefix

nolar:
I want to use simplified version tags in my projects: either "1.2.3rc" or "v1.2.3rc" (as in gevent), i.e. with no special long prefixes, such as "mylibrary-".

In current versioneer implementation, it would conflict with other tags that start with "v" (if prefix is set to "v") or with all tags (if prefix is an empty string).

Obviously, versions that start with non-digit either are very rare or do no exist at all; versions always start with a number.

So, by setting the prefix to "v", I expect that only tags like "v1.2.3rc" will be used, while tags like "vararg-experiment" or "vasilyev-patch" would not be treated as version tags (assuming that no other helper tags look like "v[digit]...").

Similarly, when I set the tag prefix to an empty string, I expect that "1.2.3rc1" would be treated as a version, while "here-it-happened-2" would not.

Now, "vasilyev-patch" leads to a version "asilyev-patch+28.gxxxxxxxx", and "here-it-happened-2" leads to a version "here-it-happened-2+28.gxxxxxxxx". And you cannot use lightweight tags for such temporary things, because "--tags" option is used in versioneer.

As tag patterns are passed to `git describe --match`, and it accepts glob(7) patterns, my suggestion is to fix this line to expect at least one digit after the prefix; all other logic with prefix removal stays the same, the digit itself goes to the version number as usually. If needed, additional option can be added for those very rare cases when version is not numeric — I can't imaging this, though, so this is not needed yet.

PS: Not sure how to add tests for this properly.

### DIFF
--- a/src/git/from_vcs.py
+++ b/src/git/from_vcs.py
@@ -26,7 +26,7 @@ def git_pieces_from_vcs(tag_prefix, root, verbose, run_command=run_command):
     # if there isn't one, this yields HEX[-dirty] (no NUM)
     describe_out = run_command(GITS, ["describe", "--tags", "--dirty",
                                       "--always", "--long",
-                                      "--match", "%s*" % tag_prefix],
+                                      "--match", "%s[[:digit:]]*" % tag_prefix],
                                cwd=root)
     # --long was added in git-1.5.5
     if describe_out is None:


nolar:
All the failed tests are due to flake8 errors, which are not part of this pull request.

effigies:
We're working on reviving this project. Can you rebase/merge master to resolve conflicts and refresh the CI?

nolar:
@effigies Hm. That was 5 years ago. I already forgot what was the problem. But I will try.

yarikoptic:
can't believe we don't have that in versioneer yet! `tag_prefix` is indeed too limiting!  I would say we better add `tag_glob` and just pass it to git describe, e.g. 

```shell
$> git describe                 
schema-0.4.0

$> git describe --match '[0-9]*'
0.2.0
```
then it would make any custom selection git can do possible.

yarikoptic:
d'oh me -- I was blind... since versioneer passes `tag_prefix` directly to `--match` isn't it just a matter of confusing name and `tag_prefix` is really a `tag_glob`? ;-)

```
$> git grep -e --match
src/git/from_vcs.py:                                     "--match", "%s*" % tag_prefix],
```
so nothing to be done beyond improving documentation? (didn't try though)

yarikoptic:
sorry for the flood -- nope. Due to trailing `*` it is still indeed a glob but wouldn't work to limit "exactly".  So adding `tag_glob` would still benefit.  As for original use-case prompted the PR it should just be possible to pass `[[:digit:]]` as part of the `tag_prefix`.

nolar:
Hello. As this PR's author, I have to admit that I do not remember why I proposed this solution, and actually what was the problem 6 years ago and if it is still the problem. 

Maybe, it was because regexps were not documented as possible values for the prefix; and it was not clear (from the docs) which type of regexps to use. The prefix implied only the strings to be used "as is". If a regexp is passed as a string prefix for an undocumented assumption about passing to `git --match`, it might eventually be broken by changing the semantics of prefixes (e.g. by escaping them) — unless documented as regexp-compatible.

Anyway, I think this PR can be closed as stale.

PR #106, opened 2015-12-14
Detect `GIT_DIR` environment variable.

skion:
This change makes versioneer work in situations where the git dir is separated from the working tree.


warner:
I think this is obsoleted by #61, which runs `git rev-parse --git-dir` instead of looking for a `.git` directory.

PR #115, opened 2016-05-09
b64decode in python3.2 accepts only bytes

grzanka:
py32 added to the test suite as special case with virtualenv < 14.0
updated README.md and setup.py with correct python versions
tox configuration modified accordingly


ant6:
@warner merge please, I need this too :)

warner:
looks good.. thanks!

PR #120, opened 2016-06-01
Extend version info by commitdate

marcelhuberfoo:
This patch adds a `date` entry to the version dictionary.
Thanks to @thurask for the initial work!

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,6 +104,9 @@ information:
 * `['full-revisionid']`: detailed revision identifier. For Git, this is the
   full SHA1 commit id, e.g. "1076c978a8d3cfc70f408fe5974aa6c092c949ac".
 
+* `['date']`: Date and time of the latest `HEAD` commit. For Git, it is the
+  commit date in ISO 8601-like format.
+
 * `['dirty']`: a boolean, True if the tree has uncommitted changes. Note that
   this is only accurate if run in a VCS checkout, otherwise it is likely to
   be False or None
--- a/src/cmdclass.py
+++ b/src/cmdclass.py
@@ -45,7 +45,7 @@ def get_cmdclass():
             print("Version: %s" % vers["version"])
             print(" full-revisionid: %s" % vers.get("full-revisionid"))
             print(" dirty: %s" % vers.get("dirty"))
-            print(" time: %s") % vers.get("time"))
+            print(" date: %s" % vers.get("date"))
             if vers["error"]:
                 print(" error: %s" % vers["error"])
     cmds["version"] = cmd_version
--- a/src/cmdclass.py
+++ b/src/cmdclass.py
@@ -45,6 +45,7 @@ def get_cmdclass():
             print("Version: %s" % vers["version"])
             print(" full-revisionid: %s" % vers.get("full-revisionid"))
             print(" dirty: %s" % vers.get("dirty"))
+            print(" time: %s") % vers.get("time"))
             if vers["error"]:
                 print(" error: %s" % vers["error"])
     cmds["version"] = cmd_version
--- a/src/from_parentdir.py
+++ b/src/from_parentdir.py
@@ -16,7 +16,7 @@ def versions_from_parentdir(parentdir_prefix, root, verbose):
         if dirname.startswith(parentdir_prefix):
             return {"version": dirname[len(parentdir_prefix):],
                     "full-revisionid": None,
-                    "dirty": False, "error": None, "time": None}
+                    "dirty": False, "error": None, "date": None}
         else:
             rootdirs.append(root)
             root = os.path.dirname(root)  # up a level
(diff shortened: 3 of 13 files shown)


warner:
Looks cool. A couple of issues:

* some places use `None` to mean "no date available", and others use an empty string. Let's standardize on using None, since that's what we do for the other keys
* so some places that do `pieces.get("date", "")` should become just `pieces.get("date")`
* what's the oldest version of git-export that supports the `$Format:ci$` syntax? I think we currently require at least git-1.5.5 (for `git describe --long`); if `ci` makes us need a newer git, we should document that somewhere, or figure out how to tolerate the git-export being too old (`date = None` is a fine fallback, just as long as it doesn't crash entirely). At the very least, let's add a comment to long_header.py (in `get_keywords()`) mentioning the oldest git that understands `%ci`.

Should we consider naming the key `date_iso8601` instead of just `date`? I'm on the fence about it. In my programs, when I'm dealing with time, I always stick to seconds-since-epoch, and leave formatting up to the UI. But here, we're getting a string (in a specific format) from git, and the most likely use case is for it to be printed as-is. And also that string contains extra timezone information that isn't included in seconds-since-epoch. So I wouldn't argue to do any parsing or conversion in versioneer. But it might be a good idea to remind callers that this is 1: a string (not a number), 2: in this specific ISO-8601 format. And also to remind future versioneer authors (e.g. those adding support for SVN or something other than git) of the format that's expected for this key, making it their responsibility to convert it if their VCS doesn't use ISO-8601 natively.

I dunno.. what do you think?

Also, I just pushed some fixes to trunk that fix flake8 on py2.6 (by downgrading the version).. could you rebase this branch and force-push, to trigger a new build, to include the py2.6 fix?

thurask:
I believe %ci format originated in [git 1.5.3](https://github.com/git/git/commit/ee8f838e0346751b21250f8d107049829b855a98), although there was a syntax change in [git 2.2.0](https://github.com/git/git/commit/466fb6742d7fb7d3e6994b2d0d8db83a8786ebcf).

warner:
Cool, thanks for the research. It looks like `%cI` (with a capital I), meaning strict-iso, is what got added in 2.2.0. And a quick test shows that unrecognized format strings are left as-is, so e.g. `$Format:%cI$` on a pre-2.2.0 git-export should be turned into `%cI`.

2.2.0 feels a bit too new to mandate (i.e. I want versioneer to keep working if you do an export with 2.1.0), but I'd like to get the strict ISO date if possible.

What do you think about declaring that git-2.2.0 or later gets you the `date_iso8601` field, and older gits will omit that field? To implement that, we'd use `$Format:%cI$` in `get_keywords()`, and if `git_get_keywords()` sees a value of "%cZ", it will pretend that it didn't see the value?

Another possibility is that we use `ci`, but `git_get_keywords()` modifies the value it gets back to be strict ISO-8601 (using the same rules that git does, which git/t/t4205-log-pretty-formats.sh says is `sed -e "s/ /T/; s/ //; s/..\$/:&/"`). That would get us `date_iso8601` all the way back to 1.5.3, which is comfortably ancient. It'd still be nice to notice un-expanded `%ci` and omit the date field.

thurask:
I'm fine with ISO 8601-like, as the ISO 8601-like format can be consistently extracted with little work from, at the very least, Git, Mercurial and Subversion.

Mercurial [>=1.0](https://selenic.com/hg/annotate/8999d1249171/mercurial/templatefilters.py):

    >hg log --limit 1 --template "{date|isodatesec}"
    2016-09-08 11:15:14 +0200

Subversion [anything after 2001](http://svn.apache.org/viewvc/subversion/trunk/subversion/include/svn_time.h?revision=839634&view=markup&pathrev=841883):

    >svn info | grep Date
    Last Changed Date: 2014-03-21 01:48:51 -0400 (Fri, 21 Mar 2014)

Git >=1.5.3:

    > git show -s --format=%ci HEAD
    2016-05-31 13:02:11 +0200

Still, if one wants to do so, I prefer converting the ISO8601-like format to actual ISO 8601 through a one liner in `git_get_keywords()` over any format voodoo:

    if line.strip().startswith("git_date ="):
        mo = re.search(r'=\s*"(.*)"', line)  # "2016-09-08 11:15:14 +0200"
        if mo:
            keywords["date"] = mo.group(1).replace(" ", "T", 1).replace(" ", "", 1)  # "2016-09-08T11:15:14+0200"

Incidentally, rebasing this PR onto your master as of yesterday (https://github.com/thurask/python-versioneer/tree/xcommit) gives [the all-clear on Travis for all versions](https://travis-ci.org/thurask/python-versioneer/builds/160745176).

warner:
Oh, right, so we can use `%ci` all the time, and convert it all the time, and get strict-8601 for anything >=1.5.3 . That sounds fine, no need to bother trying to use `%cI` or surviving if it doesn't exist.

(I'd like to be able to say that the `["date"]` piece is strict ISO-8601, I think that will help folks downstream)

warner:
I've added a patch to fix up the date that way, and updated the docs to mention that we're providing strict ISO-8601 dates. I also changed the `pieces` dict to make `["date"]` be None if we couldn't get one, rather than an empty string.

Landing (in rebased form) in 4e320bf264b momentarily. Thanks!

PR #125, opened 2016-09-16
Remove unused import in _version.py from file

akarnani:
The `_version.py` file generated for a from_file environment contained an unused `import sys` which occasionally could cause linting failures.


warner:
Looks good.. thanks! I was running into this problem myself just a few days ago.

I think that travis failure was because flake8 has stopped supporting py2.6 . I've updated the travis config to downgrade flake8 on older pythons, so I think it should pass when I merge this.

PR #139, opened 2017-01-02
Fix cmdclass, allow passing in of base cmdclass

### LINKED ISSUE #128: cmdclass clash
Versioneer creates a `cmdclass` and requires that the `setup` method uses that. While this works for the majority of packages, it breaks when a package must use a different `cmdclass`. For example numpy defines a [cmdclass](https://github.com/numpy/numpy/blob/6bb5a22ac873a76f4e80988469fe84c1067a7d58/numpy/distutils/core.py#L32). This `cmdclass` is also used by scipy and definitely relied upon by a bunch of other [scikits](http://scikits.appspot.com/scikits) -- those that follow the format specified by the [scikit-example](https://github.com/pv/scikit-example).

Specifically, the issue arises from 2 classes that Versioneer extends to create the `cmdclass` i.e `build_py` and `sdist`.

A solution would be to let `get_cmdclass()` take arguments for these base classes. i.e. having the a  signature like the following:

```python
get_cmdclass(build_py=None, sdist=None)
```
Such a change should be backwards compatible.

has2k1:
This commit makes it possible to use Versioneer with
packages that use a special `cmdclass`.

Fixes #128

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -97,6 +97,12 @@ To versioneer-enable your project:
         version=versioneer.get_version(),
         cmdclass=versioneer.get_cmdclass(),
 
+  If your project uses a special `cmdclass`, pass that `cmdclass` as
+  a parameter. For example:
+
+        from numpy.distutils.core import numpy_cmdclass
+        cmdclass=versioneer.get_cmdclass(numpy_cmdclass),
+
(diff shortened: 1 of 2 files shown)


warner:
looks good! landing momentarily.

PR #148, opened 2017-03-30
Render the badges

philschatz:
This renders the badges when viewing the `README.md` on GitHub.

You can click "Display the Rich Diff" below (it's a Page icon at the top-right of the diff below) or [click here](https://github.com/philschatz/python-versioneer/blob/patch-1/README.md)

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,12 +6,8 @@ The Versioneer
 * Brian Warner
 * License: Public Domain
 * Compatible With: python2.6, 2.7, 3.2, 3.3, 3.4, 3.5, 3.6, and pypy
-* [![Latest Version]
-(https://pypip.in/version/versioneer/badge.svg?style=flat)
-](https://pypi.python.org/pypi/versioneer/)
-* [![Build Status]
-(https://travis-ci.org/warner/python-versioneer.png?branch=master)
-](https://travis-ci.org/warner/python-versioneer)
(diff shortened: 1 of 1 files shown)


warner:
Looks great.. thanks!

warner:
oops, I tried the "squash and merge" button for the first time, and it landed an identical copy of your patch but without actually including your patch. I should have just used the regular "merge" button. Sorry about that.

warner:
Oops, I merged too early.. travis fails: https://travis-ci.org/warner/python-versioneer/jobs/216868999

I forgot that flake8 requires shorter lines, and having all the badge components on a single line violates the <=79 -character limit. The issue is that we copy the whole README into `versioneer.py`, and we want that to conform to flake8's requirements so that downstream projects can remain entirely flake8-clean. I think the best fix at this point is to stop copying all of the README into versioneer.py, maybe just have a pointer to the github repo's README URL.

philschatz:
Oh actually, the badges could be written like this:

```
[![gh-board][kanban-image]][kanban-url]
...

... (at the bottom of README.md)
[kanban-image]: https://img.shields.io/github/issues/philschatz/octokat.js.svg?label=kanban%20board%20%28gh-board%29
[kanban-url]: http://philschatz.com/gh-board/#/r/philschatz:octokat.js
```

Example:
https://github.com/philschatz/octokat.js/blob/master/README.md

PR #158, opened 2017-06-13
Use tox-travis

jayvdb:
This first version sits on top of https://github.com/warner/python-versioneer/pull/157 , so it has a green build.

[tox-travis](https://github.com/ryanhiebert/tox-travis) sets `TOXENV` and manages the tox dependencies, such as `virtualenv`, for Python 3.2 automatically.  (Note I am an occasionally contributor to `tox-travis`.)

This implicitly fixes one of the problems with the pypy3 builds, but isnt fixed entirely.

I must say that the 'trick' of adding prefix `py` to all tox testenv names is new to me, and I quite like it as quite inventive, but IMO it is better to "bend" Travis to work with tox standard tox testenv names, and tox-travis also provides a project where tox users can collaborate on problems wrt getting tox to work neatly with Travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,10 +17,9 @@ python:
   - "pypy"
   - "pypy3"
 install:
-# virtualenv>=14.0.0 is incompatible with python 3.2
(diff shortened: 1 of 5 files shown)


warner:
Looks good.. I was unaware of tox-travis, I'll have to give it a spin in my other projects. Thanks!

PR #160, opened 2017-07-26
Use absolute imports to avoid name collisions

mithrandi:
A package might have a submodule named json.


warner:
oh! good catch.

mithrandi:
All the tests are failing :( seems to be due to linter errors, though, not actually this change.

warner:
yeah, that should be fixed on trunk now, I'll merge it

PR #171, opened 2018-01-31
Added 'bdist_ext' cmdclass implementation

### LINKED ISSUE #170: Wheel packages with extensions do not get static version file
For packages containing native code extensions, versioneer will not bundle a static version file when creating a wheel. `sdist` distributions do get static file.

For example:

```
wget https://files.pythonhosted.org/packages/c6/76/cb7ac54f89dfb0a43dab5f15e7a7cfffc7755379af53bd833b731cca79df/ssh2_python-0.9.0-cp27-cp27m-manylinux1_x86_64.whl
```

`ssh2/_version.py`  in the wheel is the git version and requires git to be installed.

The [source distribution package](https://files.pythonhosted.org/packages/b4/fe/59da3f4428ffc614505f7df119225fe9f0dd2e998007a7d7fe843f861525/ssh2-python-0.9.0.tar.gz) on the other hand has the static version.

pkittenis:
For modules with native code extensions, versioneer needs a `build_ext` command class implementation to support wheels and other binary distributions.

Added integration tests for module with extension for setuptools and pip.

Resolves #170


pkittenis:
The `_editable` and `_develop` test failures look unrelated and also happen on master branch.

pkittenis:
Hi @s0undt3ch - not sure I understand what you mean. There have been no changes to `get_cmdclass`. It is used only in tests, [same way as in other tests](https://github.com/warner/python-versioneer/blob/master/test/demoapp2-setuptools/setup.py#L4).

Can you please clarify what changes are needed?

s0undt3ch:
https://github.com/warner/python-versioneer/blob/master/src/cmdclass.py#L10

Specifically https://github.com/warner/python-versioneer/blob/master/src/cmdclass.py#L73

pkittenis:
Oh right, sure, thank you.

pkittenis:
Changes requested have been made - ready for review.

effigies:
Hi @pkittenis. We're working on reviving this project. Can you rebase/merge `master` to get the CI working?

pkittenis:
* Updated for latest master.
* Updated cython source for py 3.8 compatibility.

pkittenis:
Merged and cleaned up changes.

effigies:
Thanks for doing this. Out of curiosity, can the `.c` file be generated on the fly, rather than including it directly in the repository? I'm not terribly familiar with cython, so this may be a dumb question.

pkittenis:
It can, but it would require `Cython` as a dependency, hence why it is included.

test/git/test_invocations.py
```diff
@@ -1,14 +1,22 @@
...
 from from_file import versions_from_file
 import common
 
+
 pyver_major = "py%d" % sys.version_info[0]
 pyver = "py%d.%d" % sys.version_info[:2]
 
+# For binary wheels with native code
+impl, impl_ver = get_abbr_impl(), get_impl_ver()
+abi, plat = get_abi_tag(), get_platform()
```
effigies, line 17: `get_platform()` now takes an `archive_root` argument, but it can be `None`.

```suggestion
abi = get_abi_tag()
try:
    plat = get_platform(None)
except TypeError:  # wheel < 0.34
    plat = get_platform()
```

test/git/test_invocations.py
```diff
@@ -1,14 +1,22 @@
...
 sys.path.insert(0, "src")
 from from_file import versions_from_file
 import common
 
+
 pyver_major = "py%d" % sys.version_info[0]
 pyver = "py%d.%d" % sys.version_info[:2]
 
+# For binary wheels with native code
+impl, impl_ver = get_abbr_impl(), get_impl_ver()
```
effigies, line 16:
```suggestion
impl, impl_ver = interpreter_name(), interpreter_version()
```

effigies (commented):
Overall this looks reasonable to me. We've got some future-proofing to worry about with `wheel`, and I've made suggestions. They require you also to add `packaging` to the `testenv`:

https://github.com/python-versioneer/python-versioneer/blob/64a606dea4db7ba776d64678d65564c3d9572d20/tox.ini#L10-L20

effigies:
> It can, but it would require `Cython` as a dependency, hence why it is included.

Fair enough. A couple KB shouldn't make a huge difference. If we're going to need to re-render it periodically, it might be worth figuring out how to get a Cython dependency into the virtualenvs.

effigies:
@pkittenis Sorry to bug you, but could you check out my suggestions? If you don't have time, I can create a PR based on this.

pkittenis:
Thanks, yes, will take a look.

pkittenis:
Re: cython - I'd like to hand write a simple extension function so this can be tested without needing cython. The use of cython was more to make it compatible with future Python versions, but a simple enough native extension should remain compatible with future C-API versions.

effigies:
Pushed a fix to the testing environment. I'm going to go ahead and merge. Thanks for this!

Issue #140, opened 2017-01-11
Source download version doesn't work without git 


j0kaso:
Maybe my problem is similar to #99. When I want to install the exported zip files without connecting it to git (user case), a version can not be found (therefore it's just 0+unknown). Would it be possible to add an information file/flag or similar automatically to the version which could be read/used by e.g setup.py install to handle this problem?
Many thx and btw - great work :)

warner:
How do you generate those zip files? If you use 'git archive' from a repo with tags, the `_version.py` file ought to get populated with a tag name (if any) and the SHA1 revision id, and the rest of versioneer should see that.

j0kaso:
I do it this way:
`git tag  -a vX.Y -m "version X.Y"`
`git push`
`git push origin --tags`
maybe I just do it the wrong way, my repository has then the right version

anntzer:
The "download zip" button on github (sadly) does not use git archive, and thus produces such zips (and they do not seem intent on changing that as the issue has been known for a while).  Perhaps you could print a warning in such cases (something similar to "The archive you are using contains no version information.  Either clone the repository, or rename the zip to include a version number (similarly to a sdist)")?

hmaarrfk:
Could you provide a facility to override the "0" in the "0+unknown" with something else? Maybe something like a fallback version?

Hopefully this would be pulled from the `.cfg` file.

skuschel:
`'0+unknown'` is really a useless description, but github still adds the correct sha to the downloaded file. So why not using the information that is there and labelling the version with the full sha instead?

xkortex:
I'm running into this issue with a submodule. This works fine:

```
$ git describe --dirty --always --tags
v0.1.0-4-g9fc5ca6
```

But `pip install .` (from submodule) or `pip install subdir/submodule` (from repo) gives `package-0+unknown`. 

What's weird is this command seems to work fine from the submodule:
```
$ python -c "import versioneer; print(versioneer.get_versions())"
{'version': 'v0.1.0+4.g9fc5ca6', 'full-revisionid': '9fc5ca6241ea03a3727f6b41011b297560c9d747', 'dirty': False, 'error': None, 'date': '2020-07-02T17:59:43-0400'}
```

PR #183, opened 2018-06-02
[WIP] Allow overwritting of the fallback version number.

hmaarrfk:
First attempt at resolving #140.

I know it is against the spirit of versioneer, but at least users will get some semblance of a version if they accidentally download the zip/tar from something that doesn't use git-archive. 

I think the general idea of giving maintainers breathing room in case they forget to update the constant.

I don't know if this is in the "spirit" of versioneer.

### DIFF
--- a/src/get_versions.py
+++ b/src/get_versions.py
@@ -82,7 +82,7 @@ def get_versions(verbose=False):
     if verbose:
         print("unable to compute version")
 
-    return {"version": "0+unknown", "full-revisionid": None,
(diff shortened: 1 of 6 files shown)


hmaarrfk:
Sorry about this, I guess I don't know where to put the `fallback_version` so it passes all tests.

I think the main question is if this kind of thing is in the spirit of your library.

If it is, I'll try a little harder to find the correct location (or maybe you can point me in the right direction).

PR #186, opened 2018-08-15
Support regex patten to search for tags

wsalmonct:
I would like to be able to use regex to search for version tags that don't have a nice prefix.

I was going to use prefix but while the git describe finds the tag you then test if the start of the string is equal to the prefix in a non regex way, so I guessed this would let me search while not braking your current behaviour.

Is this something you would consider merging?


effigies:
Can you merge/rebase `master`?

PR #190, opened 2018-10-27
Fix tests and deprecation warnings

moggers87:
Tests are currently broken in master, this PR fixes that.

Most of the work was making sure that installed packages used for testing were old enough to work with EOL'd versions of Python.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,9 @@ python:
   - "pypy3"
 install:
   - pip install "tox<3.0.0" tox-travis
+  # work around for travis weirdness on Python 3.3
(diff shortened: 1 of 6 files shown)


BoboTiG:
Would be a good thing to merge this one as it will resolve `DeprecationWarning: invalid escape sequence` in the generated `versioneer.py`.

alimcmaster1:
@warner is it possible to take a look at this?

effigies:
Closed and reopened to check whether this still fixes Travis...

effigies:
Alas, things are failing. If you care to have another go at this, please do. Otherwise I'll have a look this next week.

PR #207, opened 2020-04-12
Remove dep warnings

alimcmaster1:
Remove Depreciation Warnings  -

```
pandas/versioneer.py:400: DeprecationWarning: The SafeConfigParser class has been renamed to ConfigParser in Python 3.2. This alias will be removed in future versions. Use ConfigParser directly instead.
      parser = configparser.SafeConfigParser()
pandas/versioneer.py:402: DeprecationWarning: This method will be removed in future versions.  Use 'parser.read_file()' instead.
      parser.readfp(f)
```

Ref - https://github.com/warner/python-versioneer/pull/190 to fix the tests on master.

cc @warner


src/header.py
```diff
@@ -69,9 +69,9 @@ def get_config_from_root(root):
     # configparser.NoOptionError (if it lacks "VCS="). See the docstring at
     # the top of versioneer.py for instructions on writing your setup.cfg .
     setup_cfg = os.path.join(root, "setup.cfg")
-    parser = configparser.SafeConfigParser()
+    parser = configparser.ConfigParser()
```
effigies, line 72:
```suggestion
    parser = ConfigParser()
```

effigies (commented):
@alimcmaster1 Can you rebase off of `master` so we can get the tests going? You'll need to drop `pypy2` testing in `.travis.yml` and `tox.ini`.

If we do want to preserve pypy2 compatibility, here's an alternative that will allow the new syntax to be used:

```Python
try:
    from configparser import ConfigParser
except ImportError:
    from ConfigParser import SafeConfigParser as ConfigParser
    ConfigParser.read_file = ConfigParser.readfp
```

And then apply the suggested change.

src/header.py
alimcmaster1: Sure will update shortly.

setup.py
```diff
@@ -188,6 +188,9 @@ def is_pure(self): return True
         "Programming Language :: Python :: 3.2",
```
alimcmaster1, line 188: We might want to drop these versions...

alimcmaster1:
@effigies 

- I dropped PyPy2 to keep it cleaner - unless you think we should support.
- Update Repo Urls
- Update setup.py to include more recent python versions than we now run CI against.

tox.ini
```diff
@@ -38,4 +41,4 @@ commands =
     python test/run_pyflakes_src.py
     pyflakes test
     flake8 git_version.py versioneer.py
-    pep8 git_version.py versioneer.py
+    pycodestyle --max-line-length=88 git_version.py versioneer.py
```
alimcmaster1, line 44: pep8 has been renamed to pycodestyle

alimcmaster1:
@effigies you able to review - whenever you have a moment

effigies:
LGTM. Thanks!

I think dropping pypy2 probably makes the most sense, as well. Not hearing any outcry, let's do this.

PR #213, opened 2020-09-13
Fix tests, deprecation warnings, test only on live Pythons

effigies:
Builds on #190.

This PR drops Python <3.6 support altogether. This may require some discussion. If we want to support older versions for a bit, I think the main things we'll need to do are to keep the `virtualenv<20` restriction (relaxing that was too big a job for this PR) and pick Travis `dist`s that support those versions of Python.

Closes #190.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,9 @@ python:
   - "pypy3"
 install:
   - pip install "tox<3.0.0" tox-travis
+  # work around for travis weirdness on Python 3.3
+  # https://github.com/aws/base64io-python/issues/4#issuecomment-397857642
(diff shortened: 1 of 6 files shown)


alimcmaster1:
I’m happy to close #207 if this super-seeds it. I had fixed up some configparser warnings

effigies:
Forgot to update the text. I wouldn't close #207 yet, as that would require dropping support for `pypy2`. Since that's still a live project, we should have a discussion about whether to continue supporting it.

effigies:
I'm going to go ahead and merge so we can start rebasing and testing other PRs.

Others should feel free to open a PR to revert these changes.

PR #216, opened 2020-09-25
Set minimum Python to 3.6

effigies:
Following up #207 and #213 with some further cleanup.

* Removing several hacks and/or `from __future__` imports that have no effect in Python 3.6+.
* Update classifiers and add `python_requires` in package metadata
* Drop pypy2 from tox

Also removed explicit `pyflakes` and `pycodestyle` from the test environment, since `flake8` was written to wrap these two. It's very unlikely ever to change that.


effigies:
In the absence of outcry, I'm going to go ahead and merge this. Python 3.5 is officially EOL.

PR #223, opened 2020-10-21
FIX: Support extensions with pypy3

effigies:
Follow-up to #171. I hadn't noticed that the pypy3 tests started failing then.

This is from a little bit of fiddling around, but we'll see if it breaks other environments.

cc @pkittenis

### DIFF
--- a/test/git/test_invocations.py
+++ b/test/git/test_invocations.py
@@ -59,6 +59,9 @@ class _Invocations(common.Common):
         with warnings.catch_warnings():
             warnings.simplefilter("ignore")
             virtualenv.create_environment(venv_dir)
+            self.run_in_venv(venv_dir, venv_dir,
+                             'pip', 'install', '-U',
+                             'pip', 'wheel', 'packaging')
         return venv_dir
 
     def run_in_venv(self, venv, workdir, command, *args):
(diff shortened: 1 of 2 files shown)


pkittenis:
Have made some changes on top of this branch to fix PyPy binary wheel tests at #224

In short, the issue was the version of `packaging` in the virtualenv was too old and was creating wrong ABI tag for PyPy 7.3.1 and above which uses a newer and incompatible with that packaging version of pip - [see ABI tag changes](https://doc.pypy.org/en/latest/release-v7.3.1.html).

Since the tests and virtualenv package versions can go out of sync and cause failures, have added updates for pip and wheel also.

effigies:
Great! Thanks for looking into this.

effigies:
Going to try fiddling on this branch again.

test/git/test_invocations.py
```diff
@@ -397,7 +401,7 @@ def make_setuptools_wheel_with_pip(self):
 
     def make_binary_wheelname(self, app):
         return "%s-2.0-%s-%s-%s.whl" % (app,
-            "".join([impl, impl_ver]), abi, plat)
+            "".join([impl, impl_ver]), abi, plat.replace("-", "_"))
```
effigies, line 404: The issue seems to be that sometimes `impl_ver` is 372.

```
FAIL: test_extension_wheel_setuptools (__main__.SetuptoolsUnpacked)
----------------------------------------------------------------------
  File "test/git/test_invocations.py", line 853, in test_extension_wheel_setuptools
    self.assertIn(created, glob(os.path.join(wheelhouse, "*.whl")))
AssertionError: '/tmp/tmpchzmptu6/demoappext-setuptools-repo/dist/demoappext-2.0-pp36-pypy3_72-linux_x86_64.whl' not found in ['/tmp/tmpchzmptu6/demoappext-setuptools-repo/dist/demoappext-2.0-pp372-pypy3_72-linux_x86_64.whl']
```

pkittenis: That's the same error. One or multiples of setuptools/pip/packaging is out of date for PyPy >=7.3.1 and causing wrong ABI tag to be generated. Will try updating them in tox - cannot reproduce locally with updated versions.

pkittenis: #224 is working now.

PR #225, opened 2020-10-21
ENH: Test and advertise support for Python 3.9

effigies:


effigies:
This includes some fixes from #223/#224, so should probably just wait on that.

PR #229, opened 2020-10-23
FIX: Improved handling of refname edge cases

effigies:
This PR covers two edge cases.

The first is a case where the `keywords` passed to `git_versions_from_keywords()` is non-empty but `refnames` is missing. IIRC this could occur in `git archive`s for hashes that had no branches or tags associated. This breaks L54:

https://github.com/python-versioneer/python-versioneer/blob/8071eaa56816837ac670d18bdc73d7de6cb60ea6/src/git/from_keywords.py#L54

Instead we reduce the check to `'refnames' in keywords`, which covers the actually erroring case, and provide a better error message.

The second is a case where you might have a tag `test_tag`. If you have a tag prefix `'v'`, this will get filtered out. If your tag prefix is `''`, then all weird tags get pulled in. So I've added a filter that requires the first character of the tag (post-prefix) to be a digit.

### DIFF
--- a/src/git/from_keywords.py
+++ b/src/git/from_keywords.py
@@ -36,8 +36,8 @@ def git_get_keywords(versionfile_abs):
 @register_vcs_handler("git", "keywords")
 def git_versions_from_keywords(keywords, tag_prefix, verbose):
     """Get version information from git keywords."""
-    if not keywords:
-        raise NotThisMethod("no keywords at all, weird")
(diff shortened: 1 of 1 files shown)


effigies:
Anybody care to review?

effigies:
This does resolve concrete issues for me, and hasn't spawned outrage. I'll go ahead and merge...

PR #232, opened 2020-11-19
Use _build_ext from cmds

BenjaminRueth:
I had to add these lines in `myproject/versioneer.py` to be able to use a custom `build_ext` that is required in a project of mine using cython. I hope that I applied the change at the right place und the `python-versioneer`.


effigies:
Looks right. Thanks!

Issue #230, opened 2020-10-23
Discussion: Fallback/override version sources


effigies:
This issue is intended to consolidate a few parallel threads and seek a unified solution.

A few cases have come up where the version information cannot be extracted from the current state of the source code, or changes after the tag should be ignored for some reason.

* Containers (e.g., Docker) - repo contents are copied, but `.git/` is missing and no substitutions can be made by `git-archive` or `setup.py` first. (https://github.com/python-versioneer/python-versioneer/issues/189#issuecomment-490213494, https://github.com/python-versioneer/python-versioneer/issues/215)
* git archive with no refnames - commit hash is available, but no branch or tag information (https://github.com/python-versioneer/python-versioneer/issues/199)
  * git archive with branch names - possibly some information recoverable depending on branch naming strategy
* Dirty repo or extra commits for version-irrelevant changes - e.g., update CI to build new wheels (#189)

There are some suggestions for fallback or overrides:

* Set an environment variable
  * `VERSIONEER_VERSION`: https://github.com/python-versioneer/python-versioneer/issues/199#issue-463634922
  * `VERSIONEER_OVERRIDE`: https://github.com/python-versioneer/python-versioneer/issues/189#issue-364842532
* Set a static file (https://github.com/python-versioneer/python-versioneer/issues/189#issuecomment-490213494)
* Set in `setup.cfg` (https://github.com/python-versioneer/python-versioneer/issues/189#issuecomment-490213494)
* `runpy` a dependency-free file in the source tree for fallback (https://github.com/nipy/nibabel/commit/389059e5421900c2d58be0b3c3c7496aecd7f723)

A few hacks are mentioned in #199, but these seem more oriented toward package maintainers working around versioneer than something that can be implemented as policy, but perhaps I'm wrong. They may, in any event, have useful ideas for an ultimate solution.

The final consideration is the user.

* The interested user who would like to create Docker images or wheels and would be happy to modify their environment, invocation or script a bit to get the desired version that they will then publish. The one thing they don't want to do is write custom modifications to versioneer that they have to keep track of.
* The user who is just fetching the package and trying to install it, and why should they have to care about versioneer? They shouldn't, but wouldn't it be nice if however they fetched the package, it had a sensible ordering in the PEP-440 scheme?

[...]

Issue #189, opened 2018-09-28
Consider adding ability to set version manually


matthew-brett:
I use versioneer a lot, it is very useful.

But - one problem I come up against fairly often, is that I want to set the version myself.  The typical use-case is that I am building wheels with my CI scripts.  I tag the repo, and set the builds going, all good.  But then a new Python version comes out, and I need to build wheels for the new version.  I change the CI scripts to build for the new version, the repository changes and so the version changes, meaning the wheels are no longer valid.

I'd really like the ability to do something like:

export VERSIONEER_OVERRIDE=0.4

which would force versioneer to pick up my preferred version.  Is that possible, do you think?

oesteban:
Another context where this would be useful is baking a package into a container. If the `.git/` folder is not copied to the container (and there are reasons not to do that), then versioneer is totally unable to reconstruct the version string.

pskopnik:
I'm looking for this functionality as well. Use case: My CI perform changes on checked-in but not-packaged files during the build process. The `dirty` tag is then included in any distribution built from the source tree.

I would prefer a static file as well. Perhaps a `fix-version` command could be added, which writes the current state into the `_version.py` file, similarly to what is happening during sdist, ...

My workaround is to first create an sdist and then build the wheel from the sdist.

effigies:
I think there's a common issue of finding some fallback version, and I've tagged the open issues that bring up different aspects of this with the [fallback](https://github.com/python-versioneer/python-versioneer/labels/fallback) label.

It would be good to try to address this with each of these use cases in mind.

effigies:
Hi @matthew-brett @oesteban @pskopnik, would you care to have a look at #234?

PR #234, opened 2020-11-20
ENH: Provide override and fallback options

effigies:
This proposes two new configuration entries:

```INI
[versioneer]
...
override_variable = VERSIONEER_OVERRIDE
fallback_tag = 1.0.0.dev0
```

`override_variable` would allow the project to specify an environment variable that, if defined, would determine the version, no questions asked.

`fallback_tag` would also allow the project to specify a fallback if all else fails besides `0+unknown`. I would use this in my projects by bumping after a release to indicate what release I'm working towards.

```Diff
-fallback_tag = 1.0.0.dev0
+fallback_tag = 1.1.0.dev0
```

The latter use case permits installations from non-tag archives `https://github.com/<org>/<project>/archive/<githash>.tar.gz`.

Suggestion for #230. Tests to come, and might need to extend these tendrils a bit deeper into the code. Thoughts welcome.


src/setupfunc.py
```diff
@@ -17,6 +17,8 @@ def do_vcs_install(): pass # --STRIP DURING BUILD
  versionfile_build = myproject/_version.py
  tag_prefix =
  parentdir_prefix = myproject-
+ override_variable = VERSIONEER_OVERRIDE
```
hroncok, line 20: Is this the default? I'd like to avoid a situation when I need to patch the project to set this before I can use it.

effigies: Since versioneer is vendored, there's no real avoiding the patch. But once this exists you could `versioneer install` a newer version and do it without committing.

I'm also okay with making `VERSIONEER_OVERRIDE` an unconfigurable variable name. There's probably not a lot of value to having it vary from project to project.

pskopnik:
LGTM, though just read the PR explanation.

I wonder if there is a way to preserve other version information as well? `'full-revisionid'` and `'date'` are important. As written in https://github.com/python-versioneer/python-versioneer/issues/189#issuecomment-585221081, my use case is building wheel and sdist after pre-processing (cythonizing), which causes the version to become "dirty'.

I assume [`build`][build] ([PEP 517][pep-517]) allows access to environment variables by the backend and `setuptools.build_meta` passes all variables on to the `setup.py` subprocesses. But I haven't checked.

[build]: https://pypa-build.readthedocs.io/en/latest/
[pep-517]: https://www.python.org/dev/peps/pep-0517/

effigies:
> I wonder if there is a way to preserve other version information as well? `'full-revisionid'` and `'date'` are important.

In principle, sure. I think it would require a refactor where we build the version dict as we can, and then either override or fallback based on the ENV and what we find. Not a heavy lift, so if those values are important when overriding, I think we could do that.

> As written in [#189 (comment)](https://github.com/python-versioneer/python-versioneer/issues/189#issuecomment-585221081), my use case is building wheel and sdist after pre-processing (cythonizing), which causes the version to become "dirty'.

Wouldn't putting Cython outputs into `.gitignore` prevent the dirty mark?

> I assume [`build`](https://pypa-build.readthedocs.io/en/latest/) ([PEP 517](https://www.python.org/dev/peps/pep-0517/)) allows access to environment variables by the backend and `setuptools.build_meta` passes all variables on to the `setup.py` subprocesses. But I haven't checked.

It looks like they mostly pass the environment through, with Python/pip variables excluded. But it would be good to test this manually for that case...

pskopnik:
For discussing this further I would have to examine my affected project in detail. I lack the capacity to do so for the time being.
When I get around to it, I will post my (updated) problem. Perhaps a solution will become clear.
Until then, this override seems like a sensible way for manually specifying a version.

---

> Wouldn't putting Cython outputs into `.gitignore` prevent the dirty mark?

Hmm, they are, I don't remember right now what exactly triggered the mark :thinking:

effigies:
Cool. From your other comment, it sounds like a `python setup.py set_version` or similar command would also work, and I don't think any resolution here would prevent us from implementing something like that.

pskopnik:
Yes, such a command may be implementable in my `setup.py` directly. If it proves useful I could open a PR to move the function into versioneer.
We will see :)

oesteban (approved):
Seems sensible to me.

src/setupfunc.py
effigies: Made `VERSIONEER_OVERRIDE` an unconfigurable variable name.

effigies:
Haven't fully wrapped my head around the best way to build tests into versioneer, so instead debugged by using it on another project: https://github.com/nipy/nibabel/pull/976

I'm not a huge fan of how the fallback version is working out, since it requires rerunning `versioneer install` every time you bump the fallback version:
https://github.com/nipy/nibabel/commit/7e11d7ca13eea2a56658f40f00e890692079d204#diff-0a3e742348fbf3493a73cd3f000eaa6a4037a0694906468e9d059672633e6a34R47

Anybody have a better notion?

chrisroat:
With this update, what is the correct recipe when creating a docker image within a source tree?  Say the following two lines were used to install the project (no .git directory present during copy, either because of .dockerignore or its a repo subdirectory)?

```
COPY . /myproject/
RUN pip install /myproject
```

effigies:
I think you might still need a `--build-arg` to make that work.

```
docker build --build-arg VERSIONEER_OVERRIDE=$( python setup.py version | grep Version: | cut -d\  -f 2 ) .
```

effigies:
Would appreciate a review on #238, with just the override component.

effigies:
I'm done trying to update versioneer. If someone wants to pick this up, they can.

PR #235, opened 2020-12-04
Add pip install versioneer step to INSTALL

barretobrock:
Doing so makes the instructions match with what's under the quick install section of the README


effigies:
Hard to argue with. Thanks!

PR #237, opened 2020-12-06
TEST: Stop testing easy_install

effigies:
`easy_install` is deprecated in every Python version we support. If anybody still depends on it, they're probably not updating to the latest Versioneer.

What do people think?


effigies:
Nobody's outraged, and this is all deprecated. Executive merge.

PR #241, opened 2020-12-21
FIX: Respect versionfile_source in __init__.py snippet

### LINKED ISSUE #123: Versioneer does not write import inside __init__.py in accordance with versionfile_source
Example:

``` ini
[versioneer]
...
versionfile_source = src/myproject/__version__.py
versionfile_build = myproject/__version__.py
...
```

With that config `versioneer install` writes file `__version__.py`, but in `__init__.py` writes:

``` python
from ._version import get_versions
__version__ = get_versions()['version']
del get_versions
```

Notice - `from ._version ...`, not `from .__version__ ...`.

### LINKED ISSUE #218: Versioneer deleting function named `get_versions()` from module
As seen in Versioneer 0.18

The Versionner install script adds the following lines to the end of the specified `__init__.py` file:
```python
from ._version import get_versions
__version__ = get_versions()['version']
del get_versions
```

However, this has the side-effect of deleting any object in the root of the module with the name `get_versions`. This caused my code to stop working when I installed Versionner, as this was already a name used by the module.

I think that something like this would be better, in that it does not affect any objects that are not already going to be modified by the code:
```python
from ._version import get_versions as __version__
__version__ = __version__()['version']
```

This eliminates the risk of namespace collisions (except for `__version__` of course), but I am not sure if this will work correctly in all use cases of Versioneer. It seems to be working correctly for a repository installed in editable mode on my system.

staticintlucas:
Fixes #123, fixes #218.

Now uses the correct module name from versionfile_source in the \_\_init\_\_.py snippet.

Since I was changing it anyway I also included the fix mentioned in #218. The snippet no longer deletes any object called `get_versions` in the root module.

### DIFF
--- a/src/setupfunc.py
+++ b/src/setupfunc.py
@@ -44,9 +44,8 @@ SAMPLE_CONFIG = """
 """
 
 INIT_PY_SNIPPET = """
-from ._version import get_versions
-__version__ = get_versions()['version']
-del get_versions
+from . import {0}
+__version__ = {0}.get_versions()['version']
 """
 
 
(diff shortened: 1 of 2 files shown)


effigies:
Thanks, and apologies for taking so long to get to this. Please have a look at https://github.com/staticintlucas/python-versioneer/pull/1. Note, I haven't tested either of these, yet...

staticintlucas:
Those changes look good to me, I've merged your PR

effigies:
Thanks!

PR #242, opened 2020-12-23
fix possible typo in build_ext

ashutoshvarma:
I don't know if it was intentional or not, but it was breaking `pip wheel` in my cython project which has `src/myproject` style project layout.
image


effigies:
I agree, thanks.

Issue #255, opened 2021-05-26
Enhance tag matching/describe: tag_match + tag_exclude


yarikoptic:
ATM tag_prefix serves at least two roles 

```shell
$> git grep -e tag_prefix | grep -e startswith -e --match
src/git/from_keywords.py:        if ref.startswith(tag_prefix):
src/git/from_vcs.py:                                     "--match", "%s*" % tag_prefix],
src/git/from_vcs.py:        if not full_tag.startswith(tag_prefix):
```
so to pass to `git describe --match` to select tags, and then to strip that prefix from the tag.  That works quite nicely for a good range of use cases, but fails in a range of others: see #105, #182 ( probably more, yet to fish), which seems to boil down

I think that we could address majority of use cases if we allow for

- `tag_match` so we get
```
src/git/from_vcs.py:                                     "--match", "%s%s*" % (tag_prefix, tag_match)],
```
to be able to more precisely match for desired tags whenever desired, but not demand stripping that away (as if now tried to provide glob in `tag_prefix` as I ignorantly suggested in #105)

- `tag_exclude` (multiline with multiple entries allowed, .strip before use) so we pass `--exclude <tag_exclude>` option(s) to git describe.

I have not analyzed yet how it would work for `src/git/from_keywords.py` or it would need to be git specific.

WDYT @warner, @effigies (doh -- we need to setup teams), @nolar, @benjwadams, @jwodder ?

effigies:
I'm failing to understand the use cases here.

#105 I get, and it looks like a sensible solution to the problem. I'm inclined to rebase that one myself and just get it moving.

I completely fail to understand the problem in #182. I think #105 would prevent v0.1 from being considered, but presumably `git describe` wouldn't be reaching as far back as `v0.1`. I've asked them to clarify...

What is the `--exclude` solving in DANDI's use-case? Could you describe it with a series of tags?

effigies:
Pushed #105 replacement to #256.

yarikoptic:
> What is the `--exclude` solving in DANDI's use-case? Could you describe it with a series of tags?

```
(git)lena:~/proj/dandi/dandischema[master]git
$> git tag
0.0.0
0.1.0
0.1.1
0.2.0
schema-0.4.0

$> git describe
schema-0.4.0

$> git describe --exclude '[^0-9]*'
0.2.0
```
ref on more discussion etc: https://github.com/dandi/dandischema/pull/19#discussion_r640049802

[...]

PR #256, opened 2021-05-26
FIX: Use only numeric versions in Git, ignore other tags with the same prefix

effigies:
Replaces #105.

### DIFF
--- a/src/git/from_vcs.py
+++ b/src/git/from_vcs.py
@@ -32,13 +32,12 @@ def git_pieces_from_vcs(tag_prefix, root, verbose, runner=run_command):
             print("Directory %s not under git control" % root)
         raise NotThisMethod("'git rev-parse --git-dir' returned error")
 
-    MATCH_ARGS = ["--match", "%s*" % tag_prefix] if tag_prefix else []
-
     # if there is a tag matching tag_prefix, this yields TAG-NUM-gHEX[-dirty]
     # if there isn't one, this yields HEX[-dirty] (no NUM)
-    describe_out, rc = runner(GITS, ["describe", "--tags", "--dirty",
-                                     "--always", "--long", *MATCH_ARGS],
-                              cwd=root)
+    describe_out, rc = runner(GITS, [
+        "describe", "--tags", "--dirty", "--always", "--long",
+        "--match", f"{tag_prefix}[[:digit:]]*"
+        ], cwd=root)
     # --long was added in git-1.5.5
     if describe_out is None:
         raise NotThisMethod("'git describe' failed")


effigies:
@yarikoptic Care to review? Regardless of how #255 turns out, this seems like it should be correct.

yarikoptic:
>  Regardless of how #255 turns out, this seems like it should be correct.

well, `tag_match` is introduced, then we would need then probably default it similarly to that `git_describe_options` to have the `[[:digit:]]` so overridden `tag_match` could replace it whenever desired.

> @yarikoptic Care to review?

I have a general reservation against this approach for probably no more (if any) of 1% of use cases where someone does have a version which doesn't start with a digit but some alpha/letter for some obnoxious reason (e.g. `lib0.1`).  I do not think that `versioneer` ATM cares for the version to be semver or alike.  If we default to such restriction without giving a way around it, that 1% of people wouldn't be able to upgrade to newer versioneer... so boils down if we want to impose on those folks? ;-)

effigies:
> I have a general reservation against this approach for probably no more (if any) of 1% of use cases where someone does have a version which doesn't start with a digit but some alpha/letter for some obnoxious reason (e.g. `lib0.1`). I do not think that `versioneer` ATM cares for the version to be semver or alike. If we default to such restriction without giving a way around it, that 1% of people wouldn't be able to upgrade to newer versioneer... so boils down if we want to impose on those folks? ;-)

Huh.

```
nibabel$ python setup.py version
running version
keywords are unexpanded, not using
got version from VCS {'version': '3.2.1+103.ge51bcb43', 'full-revisionid': 'e51bcb43d9c6f5ad329ffb230afda0c26b9e8617', 'dirty': False, 'error': None, 'date': '2021-05-21T07:55:20-0400'}
Version: 3.2.1+103.ge51bcb43
 full-revisionid: e51bcb43d9c6f5ad329ffb230afda0c26b9e8617
 dirty: False
 date: 2021-05-21T07:55:20-0400

nibabel$ git tag -a -m "lib0.1" lib0.1

nibabel$ python setup.py version
/home/chris/miniconda3/lib/python3.8/site-packages/setuptools/dist.py:466: UserWarning: The version specified ('lib0.1') is an invalid version, this may not work as expected with newer versions of setuptools, pip, and PyPI. Please see PEP 440 for more details.
  warnings.warn(
running version
keywords are unexpanded, not using
got version from VCS {'version': 'lib0.1', 'full-revisionid': 'e51bcb43d9c6f5ad329ffb230afda0c26b9e8617', 'dirty': False, 'error': None, 'date': '2021-05-21T07:55:20-0400'}
Version: lib0.1
 full-revisionid: e51bcb43d9c6f5ad329ffb230afda0c26b9e8617
 dirty: False
 date: 2021-05-21T07:55:20-0400
```

I guess technically it supports non-PEP440 tags. Though if I try to `pip install`, it complains loudly and falls back to a deprecated method.

```
[...]
  WARNING: Built wheel for nibabel is invalid: Metadata 1.2 mandates PEP 440 version, but 'lib0.1' is not
Failed to build nibabel
Installing collected packages: nibabel
  Attempting uninstall: nibabel
    Found existing installation: nibabel lib0.1
    Uninstalling nibabel-lib0.1:
      Successfully uninstalled nibabel-lib0.1
    Running setup.py install for nibabel ... done
  DEPRECATION: nibabel was installed using the legacy 'setup.py install' method, because a wheel could not be built for it. A possible replacement is to fix the wheel build issue reported above. You can find discussion regarding this at https://github.com/pypa/pip/issues/8368.
Successfully installed nibabel-lib0.1
```

It feels to me like not something we should try to support.

yarikoptic:
> I guess technically it supports non-PEP440 tags. Though if I try to pip install, it complains loudly and falls back to a deprecated method.

yes it does.  Do we know if versioneer is only used for Python projects?  anyways -- I do agree that it feels that we shouldn't even bother supporting versions which do not start with version, but 

1. versioneer does support them ATM so these change has potential to break existing workflows. Somewhat good that versioneer is not used as a module and usually people just carry a copy, right?  Only "somewhat" is that we wouldn't discover if any workflow got broken until awhile ;)
2. we don't know if/how many people use it for some odd (or not so odd, just rare) scenarios where they might like some adhoc version to be populated (possibly to be tuned up later "in code").

effigies:
It is specifically a patch into distutils/setuptools' `setup.py` implementation, so the 99+% use case is going to be Python packages. Maybe somebody's doing wacky things with `setup.py` to create non-Python packages, but a simple solution to the 99+% case seems better than a complex solution to a niche use case that we have no evidence for.

As you say, people do need to manually upgrade to a new version of versioneer, and we can get a report when it happens. Your approach isn't going to be harder to implement at that time, and that time may never come.

effigies:
Supporting non-python projects is not in scope, so non-PEP-440 tags are not in scope.

PR #261, opened 2021-08-22
ENH: Respect tags with ``.postN`` in pep440-pre style

nightlark:
tl;dr - `pep440-pre` doesn't check if the closest tag already contains a post release segment, resulting in a version string that PyPI rejects because there are multiple post release segments. The proposed change detects if a post release segment is present in the tag, and makes versioneer create dev releases for the next postN version instead of always post0.

---

For a number of projects (e.g. cmake, ninja), the main release segment of the version string matches the version number of the tool being packaged, and the post-release segment is used to indicate updates that just change the pip package but not the underlying tool.

Picking a style to use with versioneer that fits the tag/version workflow is tricky because none of the ones that append a `+<local version>` string can be uploaded to PyPI. The closest match to this versioning scheme is `pep440-pre`, but because new releases use tags in the form `vX.Y.Z.postN`, versioneer decides the version will be in the form `vX.Y.Z.postN.post0.devN`, which PyPI rejects because there are multiple post-release segments.

In addition, if the last release was `vX.Y.Z.post0`, the next post release would be `vX.Y.Z.post1`, so versioneer dev versions should be in the form `vX.Y.Z.post1.devN`. (This has the downside of if the next release won't be a `post` release -- but that same downside is true of the current implementation of the `pep440-pre` style.)

I've modified the `pep440-pre` style to check if a post release segment exists in the tag, and if so to increment the `postN` part of the rendered version to the appropriate value for the "next" post release. I think this handles the cases described above better than the existing `pep440-pre` implementation without breaking backwards compatibility. The one downside with what I have right now is it expects that tag to be the "normalized" form of the tag in that it doesn't handle synonyms like `rev`, `r`, or the alternative `-`, `_`, no segment separator, or `-` only variations of specifying a post release.

Any thoughts on making these changes to the existing style, or adding a new `pep440-pre-noduplicatepost` style? I think there are a few other of the `pep440` styles that could also be updated to better handle a closest version tag that already includes pre/post/dev release segments.


src/render.py
```diff
@@ -61,16 +61,29 @@ def render_pep440_branch(pieces):
...
 
 
+def pep440_split_post(ver):
+    """Split pep440 version string at the post-release segment.
+
+    Returns the release segments before the post-release and the
+    post-release version number (or -1 if no post-release segment is present).
+    """
+    vc = str.split(ver, ".post")
+    return vc[0], int(vc[1] or 0) if len(vc) == 2 else -1
```
nightlark, line 71: The else case should probably return type `None` instead.

effigies:
I think I'm the closest person. I'll try to have a look soonish.

nightlark:
There are a few packages I've been holding off sending PRs to for using this modified version of the renderer -- if this PR seems like it is moving somewhere soon then I'll continue to hold off, since I'd much rather the changes get upstreamed to versioneer rather than maintaining my own fork.

effigies (approved):
Thanks for the bump. This looks entirely reasonable to me. I don't have a strong opinion about the `-1` vs `None` issue. If you would prefer to change it, feel free. Otherwise I think this is good to merge.

nightlark:
Thanks! I'll update the docs to mention the new behavior. I think I'll make the `None` change too, since when looking at it again I realized that if the distance from the last tag is 0 it shouldn't increment the post release segment.

src/render.py
```diff
@@ -61,16 +61,29 @@ def render_pep440_branch(pieces):
...
     Exceptions:
     1: no tags. 0.post0.devDISTANCE
     """
     if pieces["closest-tag"]:
-        rendered = pieces["closest-tag"]
-        if pieces["distance"]:
-            rendered += ".post0.dev%d" % pieces["distance"]
+        tag_version, post_version = pep440_split_post(pieces["closest-tag"])
+        rendered = tag_version
+        if post_version > -1 or pieces["distance"]:
```
nightlark, line 83: This should be an `and`, if there have been no commits the closest-tag should be used as is, otherwise the pre-release dev segment and incremented post release segment need adding.

effigies:
LGTM. Happy to merge if you're ready.

Issue #122, opened 2016-08-16
--match <tag> causes "--dirty is incompatible with commit-ishes"


msarahan:
This seems to only affect me on Windows.  Versioneer v.0.16.

At https://github.com/warner/python-versioneer/blob/master/src/git/from_vcs.py#L32 - the addition of the --match argument with the tag prefix causes versioneer to fail completely and return the version as 0+unknown.  Removing these two arguments makes things work again.

Does this warrant some platform special casing?  I'm happy to put a PR forward, but I don't have a feeling for the best way to address this issue.

alexwidener:
Seconded.

warner:
Huh. what version of git are you using? I'd expect `--match` success or failure to depend upon git, more than on the platform.

Maybe we could change it to watch for a return value that indicates a `--match` -less git (or OS) and fall back to some other flavor of the command. Or run two separate git commands (one with `--match`, one without) and merge their results.

alexwidener:
Apologies, I will be unable to contribute to this issue, I filed this when I was at a different job. I recall it happening moderately frequently, as we used Versioneer in our package management system, but I'm not currently in an environment that uses Versioneer in our package management system so I'll be unable to perform any further testing.

warner:
No worries.. thanks for confirming it happens in more than one place!

henniss:
I can confirm this. git 2.14.0, not that it matters. 

I'm getting a message `fatal: --dirty is incompatible with commit-ishes` when running `python setup.py bdist_wheel`.  It only occurs with an empty `tag_prefix`. 

It looks like the `*` is being globbed on windows, which results in multiple arguments following `--match`. For example, `subprocess.Popen(['echo', '*'])` will result in one argument on unix systems, but many arguments on windows. 

I'll try to dig into this more tonight.

drewejohnson:
I too am seeing this issue with versioneer v0.18, but only when using my anaconda distribution of python through cygwin. If I use python 3.6 or 2.7 through cygwin, `python setup.py version` yields the correct version tag. With the anaconda distribution [python 3.5.4], I get the `--dirty is incompatible with commit-ishes` error. I'm using git 2.14.1 as well

effigies:
I'm curious if this is resolved by #229.

DancingQuanta:
Hi, after some analysis of `versioneer.py` I came up with a fix in my PR #262. Could you try this out please?

PR #262, opened 2021-08-23
FIX: Escape asterisk in ``git describe`` call on Windows

DancingQuanta:
I have created a fix for #122.

This fix makes use of an existing check for `win32` to modifies the argument to `--match` accordingly so that the `*` is escaped for Windows users.

### DIFF
--- a/src/git/from_vcs.py
+++ b/src/git/from_vcs.py
@@ -22,8 +22,10 @@ def git_pieces_from_vcs(tag_prefix, root, verbose, runner=run_command):
     version string, meaning we're inside a checked out source tree.
     """
     GITS = ["git"]
+    TAG_PREFIX_REGEX = "*"
     if sys.platform == "win32":
         GITS = ["git.cmd", "git.exe"]
+        TAG_PREFIX_REGEX = "\*"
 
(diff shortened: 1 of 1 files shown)


src/git/from_vcs.py
```diff
@@ -22,8 +22,10 @@ def git_pieces_from_vcs(tag_prefix, root, verbose, runner=run_command):
     version string, meaning we're inside a checked out source tree.
     """
     GITS = ["git"]
+    TAG_PREFIX_REGEX = "*"
     if sys.platform == "win32":
         GITS = ["git.cmd", "git.exe"]
+        TAG_PREFIX_REGEX = "\*" # noqa: W605
```
effigies, line 28: Should just escape the backslash, rather than shield it from QA checks.

```suggestion
        TAG_PREFIX_REGEX = "\\*"
```

Or you could do:

```suggestion
        TAG_PREFIX_REGEX = r"\*"
```

DancingQuanta: I am trying to make sure that linux tests passes. The `flakes` does not like the string `"\*"`. I am trying to make sure that the whole string is passed to command invocation. The raw string is a good idea I didn't think of.

PR #263, opened 2021-08-25
CI: Run tests on Windows

effigies:
Inspired by #262. Let's see if we can get tests going on Windows.

### DIFF
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -19,7 +19,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: ['ubuntu-latest']
+        os: ['ubuntu-latest', 'windows-latest']
         python-version: [3.6, 3.7, 3.8, 3.9, 'pypy-3.7']
 
     steps:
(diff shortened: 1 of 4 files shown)


DancingQuanta:
I have spent some time reading through the tests, and it appears that the virtualenv setup is the first issue to tackle in testing suite. The virtualenv invocation is only designed for Linux, and so I need to understand what this step actually do.
One indication is the paths used. These paths need to be adapted to windows' style paths.
One way to do this is to create another class from `_Invocations` and use windows path in methods or other way of using paths.
I am not experienced in automating creation of virtualenvs and so this will take me some time to work out what is the right step forwards.

DancingQuanta:
Ca I have advice on `_Invocation` please? Is this class the original product of `python-versioneer`? Or was this class inspired by other projects? If there are many projects using similar techniques, there might to be one with a solution for Windows.

effigies:
As far as I know, it was originally written for Versioneer. I haven't seen anything like it elsewhere.

effigies:
Closing in favor of #283.

PR #264, opened 2021-09-02
MAINT: EnvironmentError is an alias of OSError

mwtoews:
A change from Python 3.3 is that several errors were merged to [OSError](https://docs.python.org/3.7/library/exceptions.html#OSError), including EnvironmentError

```python
>>> EnvironmentError is OSError  # True
```

so this PR simply replaces EnvironmentError with OSError.


effigies:
LGTM. Thanks!

PR #266, opened 2021-09-17
Fix issue (?) found by codespell

DimitriPapadopoulos:
codespell complains:
```
./NOTES:60: delink ==> unlink
```

### DIFF
--- a/NOTES
+++ b/NOTES
@@ -57,7 +57,7 @@
         - good to know: make_release_tree() creates hardlinks if possible. If
           'setup.py build' does this too, that would explain why it's not
           always necessary to rebuild when source files are changed.
-        - need to delink _version.py before changing it
+        - need to unlink _version.py before changing it
         - make_release_tree(base_dir, files) creates os.path.join(base_dir,f)
           for f in files
 **** let's investigate this:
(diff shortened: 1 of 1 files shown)


effigies:
Reading this, I believe it means that the file should still exist, but shouldn't be a hardlink or symlink to another file. I think "unlink" was specifically avoided for this reason.

DimitriPapadopoulos:
I see. Should this perhaps be modified to something like the following?
_make sure _version.py is not a symlink or hardlink before changing it_

effigies:
Would codespell accept quotes around it, like `"delink"`? Or what about "break the link"?

DimitriPapadopoulos:
I think codespell doesn't  take punctuation into account. In any case don't bother about codespel, I can add a config file to shut it up if needed (unfortunately [`tox.ini` won't work so it'll be `setup.cfg` or `.codespellrc`](https://github.com/codespell-project/codespell#using-a-config-file)).

NOTES
```diff
@@ -57,7 +57,7 @@
         - good to know: make_release_tree() creates hardlinks if possible. If
           'setup.py build' does this too, that would explain why it's not
           always necessary to rebuild when source files are changed.
-        - need to delink _version.py before changing it
+        - need to unlink _version.py before changing it
```
effigies, line 60:
```suggestion
        - need to ensure _version.py is not a link before changing it
```

effigies:
Sure. Made a suggestion.

DimitriPapadopoulos:
Perfect! Thank you.

PR #267, opened 2021-09-19
Either os.path or path - not both

DimitriPapadopoulos:
Use either:
```python
	import os
	os.path...
```
or:
```python
	from os import path
	path...
```

Don't use both! I have changed evrything to the latter because it was used much more often, but using the former would have the benefit of working around an LGTM.com recommendation:
https://lgtm.com/projects/g/python-versioneer/python-versioneer/snapshot/0fac2b06f40206face530fc992d3978642daece5/files/setup.py?sort=name&dir=ASC&mode=heatmap#x1c1370788a826155:1

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,7 +1,6 @@
 #!/usr/bin/env python
 
 import os, base64, tempfile, io
-from os import path
 from pathlib import Path
 from setuptools import setup, Command
 from setuptools.command.build_py import build_py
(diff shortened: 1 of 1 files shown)


effigies:
I think I would prefer `os.path` to `from os import path`, and to use `pathlib.Path` when it's cleaner. Okay to patch your PR, or would you prefer to do it?

DimitriPapadopoulos:
Please do. I could do the `os.path` → `from os import path` part, but I'm not certain what to do about the `pathlib.Path` part - it's too recent for someone who started with Python 1 :smile:

DimitriPapadopoulos:
Excellent, thank you.

Issue #268, opened 2021-10-05
mypy fails at _version.py


headtr1ck:
After installing versioneer v0.20 in our project `mypy` fails with:
```
package/_version.py:54: error: Need type annotation for "LONG_VERSION_PY" (hint: "LONG_VERSION_PY: Dict[<type>, <type>] = ...")
```
I know that one could simply ignore this file in the mypy config, but since it is only this single complaint you could maybe add the type hint? If I see it correctly it is simply:
```
LONG_VERSION_PY: dict[str, str] = {}
```

effigies:
I think that's correct, and I'm fine with it if it doesn't break Python 3.6. Care to make a PR?

PR #269, opened 2021-10-05
FIX: mypy fails after install

headtr1ck:
Fix for https://github.com/python-versioneer/python-versioneer/issues/268

Simply add type hints to `LONG_VERSION_PY` (use typing classes to support python<3.9)

### DIFF
--- a/src/git/long_header.py
+++ b/src/git/long_header.py
@@ -14,6 +14,7 @@ import os
 import re
 import subprocess
 import sys
+from typing import Callable, Dict
 
 
 def get_keywords():
@@ -51,8 +52,8 @@ class NotThisMethod(Exception):
     """Exception raised if a method is not valid for the current scenario."""
 
 
-LONG_VERSION_PY = {}
-HANDLERS = {}
+LONG_VERSION_PY: Dict[str, str] = {}
+HANDLERS: Dict[str, Dict[str, Callable]] = {}
 
 
 def register_vcs_handler(vcs, method):  # decorator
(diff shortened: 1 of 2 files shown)


effigies:
I assume you've verified that this change resolves your issue?

headtr1ck:
Yes, I have tested it and mypy no longer complains.

It could be also possible to add tests to check the compatibility in the future.
Or even check mypy for the whole repo.

headtr1ck:
It seems that setuptools is lazy with type hints, so the verioneer.py file causes issues.
However, most projects will only check the src directory and not the root directory, so only checking the git_version.py is fine.

headtr1ck:
Any idea why the setup fails?

effigies:
mypy does not seem to install on pypy (https://github.com/python/typed_ast/issues/111). I don't know if Tox will allow us to skip certain dependencies and test steps. Feel free to look into it, or I can try to find time this weekend.

tox.ini
```diff
@@ -27,6 +27,7 @@ deps =
     virtualenv<20
     packaging>=20
     pip>=20
+    mypy
```
effigies, line 30:
```suggestion
    !pypy3: mypy
```

tox.ini
```diff
@@ -47,3 +48,4 @@ commands =
     pyflakes test
     flake8 git_version.py versioneer.py
     pycodestyle --max-line-length=88 git_version.py versioneer.py
+    mypy git_version.py
```
effigies, line 51:
```suggestion
    !pypy3: mypy git_version.py
```

effigies (commented):
Think I found the trick...

PR #274, opened 2021-10-22
FIX: Escape asterisk for Windows only when tag_prefix is empty

### LINKED ISSUE #273: Tag prefix broken on Windows with versioneer 0.21
https://github.com/python-versioneer/python-versioneer/pull/262 contains a fix for handling * on windows unfortunately this breaks detecting the correct version for me.

In our repository https://github.com/qcodes/qcodes all tags are prefixed with `v` and we have `tag_prefix = v` this used to work fine however with the latest release of versioneer we get versions reported as unknown + x

Looking at the generated `git describe` command we see 

```
git describe --tags --always --long --dirty --match v*
v0.29.0-288-g9f93a1286
```

```
git describe --tags --always --long --dirty --match v\*
9f93a1286
```

```
git describe --tags --always --long --dirty --match v\\*
9f93a1286
```

so it would seem like the escape of * does not always work correctly.
removing the escaped prefix for windows restores normal functionality for us

effigies:
Fixes #273.

@DancingQuanta @mhkline @jenshnielsen Can you test to ensure this covers both uses cases? Unfortunately we haven't gotten CI tests working on Windows, so I can't test, myself.

### DIFF
--- a/src/git/from_vcs.py
+++ b/src/git/from_vcs.py
@@ -40,7 +40,8 @@ def git_pieces_from_vcs(tag_prefix, root, verbose, runner=run_command):
     # if there isn't one, this yields HEX[-dirty] (no NUM)
     describe_out, rc = runner(
         GITS,
-        [["describe", "--tags", "--dirty", "--always", "--long", "--match", match_pattern]
+        [["describe", "--tags", "--dirty", "--always", "--long",
+          "--match", match_pattern]
          for match_pattern in match_patterns],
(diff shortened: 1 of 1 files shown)


hwjohn:
If the `tag_prefix` is an empty string, I think you could just omit the `--match` argument altogether.  This would avoid having to guess whether to escape the asterisk or not.

effigies:
Closing in favor of #283.

Issue #275, opened 2021-10-25
Stop using distutils


jwodder:
This package uses distutils in several places.  However, [distutils was officially deprecated in Python 3.10 and is currently slated for complete removal from the standard library in 3.12.](https://www.python.org/dev/peps/pep-0632/)  The proper replacement is setuptools.

PR #276, opened 2021-10-25
Switch selected uses of distutils to setuptools

jwodder:
See #275.

Note that this PR does not completely remove all uses of distutils:

* Imports that use setuptools if available and then fall back to distutils if it's not are kept, as they should be harmless going forwards.
* The import of `distutils.command.build_scripts.build_scripts` in `test/demoapp-script-only/setup.py` is kept, along with the reference to `distutils.command.build_scripts` in `INSTALL.md`, as setuptools proper does not seem to have an equivalent class.  You may want to ask on [pypa/setuptools](https://github.com/pypa/setuptools) about the preferred way to update this import.

### DIFF
--- a/src/cmdclass.py
+++ b/src/cmdclass.py
@@ -31,7 +31,7 @@ def get_cmdclass(cmdclass=None):
     cmds = {} if cmdclass is None else cmdclass.copy()
 
     # we add "version" to both distutils and setuptools
-    from distutils.core import Command
+    from setuptools import Command
 
     class cmd_version(Command):
         description = "report generated version string"
--- a/src/cmdclass.py
+++ b/src/cmdclass.py
@@ -31,7 +31,10 @@ def get_cmdclass(cmdclass=None):
     cmds = {} if cmdclass is None else cmdclass.copy()
 
     # we add "version" to both distutils and setuptools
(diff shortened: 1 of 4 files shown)


src/cmdclass.py
```diff
@@ -31,7 +31,7 @@ def get_cmdclass(cmdclass=None):
     cmds = {} if cmdclass is None else cmdclass.copy()
 
     # we add "version" to both distutils and setuptools
-    from distutils.core import Command
+    from setuptools import Command
```
effigies, line 34: IIUC setuptools monkey-patches `distutils.core`? I think to maintain current behavior while bypassing the deprecation:

```suggestion
    try:
        from setuptools import Command
    except ImportError:
        from distutils.core import Command
```

test/demoapp2-distutils-subproject/subproject/setup.py
```diff
@@ -1,5 +1,5 @@
 
-from distutils.core import setup
+from setuptools import setup
```
effigies, line 2: This and the other is specifically a distutils test.

jwodder:
@effigies Review comments addressed.

PR #277, opened 2021-12-03
Format with black

EwoutH:
When updating python-versioneer on Geopandas our linters when off because [black](https://github.com/psf/black) formatting wasn't followed. So I reformatted the whole repository with black 21.11b1.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -15,6 +15,7 @@ VERSION = "0.22.dev0"
 def ver(s):
     return s.replace("@VERSIONEER-VERSION@", VERSION)
 
+
 def get(fn, add_ver=False, unquote=False, do_strip=False, do_readme=False):
     with open(fn) as f:
         text = f.read()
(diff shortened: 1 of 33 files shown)


EwoutH:
I also created a CI job that check if black formatting is followed: #278

effigies:
I would vote against this change. It can't satisfy every black configuration (as line length is a configurable parameter). You can exclude versioneer files from black in your pyproject.toml.

EwoutH:
Of course it won’t satisfy every configuration, but I still think it’s a useful change to follow the default black formatting. It makes quotation markers, white spaces and multi-line formatting consistent.

hmaarrfk:
Just my 2 cents:

I think that for the files that are required to be vendored, this would decrease the barrier to adoption.

People typically relax black formatting parameters, so limiting a vendored file to 79 characters would make it compatible with most repos. 

The code within versioneer is less important since user's aren't vendoring it.

effigies:
On the whole, I'm still fairly negative on this, but if it's important to people to have versioneer conform to default black, I'm willing to consider it. We need:

1) Not to break the repository. Note that just running `black` on the whole repository has added in many extra lines and split up lines that were supposed to be removed (tagged with `# --STRIP DURING BUILD`) so that they no longer will be. You can probably use `sty:off`/`sty:on` to get a handle on this.
2) Limit `black` to `src/` for now. Changing `setup.py` and `test/` is beyond the point at hand and just adds more things to look through.
3) Run `black --check` on `versioneer.py` and `git_version.py` in `tox`. If running black on `src/` doesn't produce black-conformant outputs, then it's going to be very hard to keep from breaking formatting in the future.

If we get a PR that satisfies these things, I'll merge it. It will probably result in a one-off release, as mixing style changes with semantic changes is painful and versioneer releases often result in PRs against many repositories.

effigies:
Created an issue for this topic.

PR #278, opened 2021-12-03
CI: Add job that checks black formatting

EwoutH:
Add job to the CI that checks if black formatting if followed


effigies:
This should be done in tox. See https://github.com/python-versioneer/python-versioneer/pull/277#issuecomment-1046274410.

PR #280, opened 2022-01-11
FIX: Unset GIT_DIR environment variable while retrieving version information from git

### LINKED ISSUE #279: Handle GIT_DIR overrides
Via https://github.com/datalad/datalad/issues/6318, when a versioneer-versioned project is used as a [remote helper](https://git-scm.com/docs/gitremote-helpers) AND installed in editable mode, you can end up with a conflict when `project.__version__` is calculated. The reason is that the `GIT_DIR` environment variable is set to the target repository, which causes versioneer's calls to `git` to look for tags in the target repository instead of the remote helper's repository.

The fix is to override with `--git-dir=.git` (all of versioneer's `git` calls have `PWD` set to the repository root).

I'm trying to figure out whether there is a case where a project might want to override this with a configuration variable. It is probably quite rare, but people setting `GIT_DIR` in their management scripts will have adjusted versioneer's behavior transparently.

effigies:
Closes #279.

~~Unclear whether there should be a config variable to allow for different `.git` locations. May not be worth the complexity to cover if the edge case is never actually hit.~~

### DIFF
--- a/src/get_versions.py
+++ b/src/get_versions.py
@@ -1,4 +1,5 @@
 import os, sys # --STRIP DURING BUILD
+from unittest import mock # --STRIP DURING BUILD
 def get_root(): pass # --STRIP DURING BUILD
 def get_config_from_root(): pass # --STRIP DURING BUILD
 def versions_from_file(): pass # --STRIP DURING BUILD
@@ -61,14 +62,17 @@ def get_versions(verbose=False):
 
     from_vcs_f = handlers.get("pieces_from_vcs")
     if from_vcs_f:
-        try:
-            pieces = from_vcs_f(cfg.tag_prefix, root, verbose)
-            ver = render(pieces, cfg.style)
-            if verbose:
-                print("got version from VCS %s" % ver)
-            return ver
-        except NotThisMethod:
-            pass
+        # Patch os.environ to allow handlers to modify their
+        # environment without affecting later subprocesses
+        with mock.patch.dict(os.environ):
+            try:
+                pieces = from_vcs_f(cfg.tag_prefix, root, verbose)
+                ver = render(pieces, cfg.style)
+                if verbose:
+                    print("got version from VCS %s" % ver)
+                return ver
+            except NotThisMethod:
+                pass
 
     try:
         if cfg.parentdir_prefix:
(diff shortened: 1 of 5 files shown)


mathijsvdv (commented):
I have reviewed your code and added the pull request effigies/python-versioneer#1 with my suggested changes. Have a look!

mih:
I tried this PR in the context of the functionality were you identified the previous fix for the interference of `GIT_DIR` with versioneer (context https://github.com/mih/datalad-mihextras/pull/9). With this change, it leads to an unconditional `fatal: KeyError('GIT_DIR')` in that implementation, because it must obtain the `GIT_DIR` setting provided by Git to the remote helper process via this environment variable.

It seems with this PR the `GIT_DIR` is removed from the process environment -- and not only for versioneer, but entirely. This breaks the environment that Git remote helpers must operate in.

The relevant change seems to be this one:

```diff
-    out, rc = run_command(GITS, ["--git-dir=.git", "rev-parse", "--git-dir"], cwd=root,
-                          hide_stderr=True)
+    # GIT_DIR can interfere with correct operation of Versioneer.
+    # It may be intended to be passed to the Versioneer-versioned project,
+    # but that should not change where we get our version from.
+    os.environ.pop("GIT_DIR", None)
+
+    _, rc = runner(GITS, ["rev-parse", "--git-dir"], cwd=root,
+                   hide_stderr=True)
```

(from diff'ing the effective change to that DataLad extension package's code after updating versioneer in it).

It would be great if the environment modification could be constraint to versioneer itself, rather than effecting the entire process.

Thx!

effigies:
Did you apply the mock as well? That is intended to protect the environment from being changed downstream.

mih:
I did not do any custom editing. I only reinstalled versioneer in the project, assuming that would be enough.

effigies:
Hmm. Okay. Let me try again...

effigies:
Missed @mathijsvdv's PR. Seems like a better approach.

effigies:
Okay, Windows tests are passing, which was the big hold up before.

@mih Can you try again with these updates?

effigies:
Hey @mih, have you had a chance to check this? If we fail again, could you share a reproduction?

mih:
Sorry for the delay. I was now able to verify that things are splendid now! Thx much for taking the time to wait for the feedback! Looking forward to the coming release.

effigies:
Thanks for testing! Will plan to release today.

PR #281, opened 2022-01-11
CI: Run weekly tests to ensure recent baseline

effigies:


effigies:
We either need to drop distutils altogether or set some max versions for the purposes of testing.

Issue #282, opened 2022-02-09
`test_dist` and `test_dist_subproject` failing


mathijsvdv:
I am currently working on implementing a fix to #273 and #263. Performing the first run of the  `tox.yml` workflow I noticed that two tests are currently failing in the master branch:
- `git/test_invocations.py::DistutilsRepo::test_sdist`
- `git/test_invocations.py::DistutilsRepo::test_sdist_subproject`

The error is the same in each case: I find that 
```python
"demoapp2-2.0/src/demoapp2.egg-info/PKG-INFO" in t.getnames()
```
actually holds true, whereas it should be false. Namely the `t.getnames()` evaluates to the following list:
```python
[
    'demoapp2-2.0',
    'demoapp2-2.0/MANIFEST.in',
    'demoapp2-2.0/PKG-INFO',
    'demoapp2-2.0/README',
    'demoapp2-2.0/bin',
    'demoapp2-2.0/bin/rundemo',
    'demoapp2-2.0/setup.cfg',
    'demoapp2-2.0/setup.py',
    'demoapp2-2.0/src',
    'demoapp2-2.0/src/demo',
    'demoapp2-2.0/src/demo/__init__.py',
    'demoapp2-2.0/src/demo/_version.py',
    'demoapp2-2.0/src/demoapp2.egg-info',
    'demoapp2-2.0/src/demoapp2.egg-info/PKG-INFO',
    'demoapp2-2.0/src/demoapp2.egg-info/SOURCES.txt',
    'demoapp2-2.0/src/demoapp2.egg-info/dependency_links.txt',
    'demoapp2-2.0/src/demoapp2.egg-info/top_level.txt',
    'demoapp2-2.0/versioneer.py'
]
```

What could have caused this?

effigies:
Distutils is being deprecated. Feel free to disable any distutils-related tests.

mathijsvdv:
Will do, thanks for the quick confirmation.

PR #283, opened 2022-02-15
Fix broken tag prefix on Windows and add CI

### LINKED ISSUE #273: Tag prefix broken on Windows with versioneer 0.21

mathijsvdv:
This pull request fixes #273 by simply removing the `--match` argument when `tag_prefix = ''`. No platform-dependent escaping of the asterisk needed.

As a first verification I tracked what `git describe --tags --match <Match string>` produces for each platform (Windows/Linux) and shell (cmd/powershell/bash), given different arguments for `--match`. In the table below we have a repository with two tags: `aaa-999` and `demo-1.0`. 

| Match string | Ubuntu Result | Windows Bash Result | Windows Cmd Result | Windows Powershell Result |
| :----------- | :------------ | :------------------ | :----------------- | :------------------------ |
| \<Leave out `--match`>    | aaa-999       | aaa-999             | aaa-999            | aaa-999                   |
| *            | fatal         | fatal               | aaa-999            | aaa-999                   |
| \*           | aaa-999       | aaa-999             | fatal              | fatal                     |
| demo-*       | demo-1.0      | demo-1.0            | demo-1.0           | demo-1.0                  |
| \demo-*      | demo-1.0      | demo-1.0            | demo-1.0           | demo-1.0                  |
| demo-\*      | demo-1.0      | demo-1.0            | fatal              | fatal                     |

The above suggests that we can indeed cover all cases by:
- Leaving out `--match` when `tag_prefix = ''`
- Using `--match {tag_prefix}*` otherwise

Next, I added the `test_no_tag_prefix` unittest to cover the case where `tag_prefix` is empty.

This covers the fix, but most of my efforts have actually been to get the Windows tests working to be sure that the implementation is correct. I started from #263 and added the following changes:
- Use `os.path.join` wherever possible rather than hardcoding forward-slash paths 570fcc8fe3083425770de87ebf8215f585e28bd5
- For some test projects, the `rundemo` file is a Python file that is made runnable in Linux through a shebang line. This doesn't work in Windows, so I opted to make a `python rundemo` call in these cases instead.
- `pip install -U pip` doesn't seem to work in Windows because of access issues. You need to use `python -m pip install -U pip` instead, so I changed this in the `run_in_venv` function.
- An environment made by `virtualenv<20` is not recognized by Windows in `python>=3.7` because it misses a `pyvenv.cfg` file. Therefore I switched to `virtualenv>=20` and made the necessary changes to the testing code.

A couple of attention points, open to suggestions here:
- As highlighted in #282, `distutils` are deprecated and the `test_sdist` and `test_sdist_subproject` were already failing in `master`, so I took the liberty of removing these two unittests. 
- The CI for Windows adds a lot of runtime: each job (6 python versions are tested each for Windows and Linux) took about 20 min on Windows, versus 4-5 min on Linux. Part of the reason is that I run the `test_git` script from bash, CMD and powershell when on Windows to cover the bases in the table above. Perhaps it's worth reducing the number of Python versions checked for Windows or to remove one of those extra scripts (CMD and Powershell seem to behave identically here).

### DIFF
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -20,7 +20,7 @@ jobs:
     strategy:
       matrix:
         os: ['ubuntu-latest']
-        python-version: [3.6, 3.7, 3.8, 3.9, 'pypy-3.7']
+        python-version: [3.6, 3.7, 3.8, 3.9, '3.10', 'pypy-3.7']
 
     steps:
       - uses: actions/checkout@v2
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -38,3 +38,11 @@ jobs:
         run: tox
         env:
           PLATFORM: ${{ matrix.os }}
+      - name: Test Git (cmd)
+        run: python test/git/test_git.py -v
+        shell: cmd
+        if: ${{ matrix.os == 'windows-latest' }}
+      - name: Test Git (Powershell)
+        run: python test/git/test_git.py -v
+        shell: pwsh
+        if: ${{ matrix.os == 'windows-latest' }}
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -12,18 +12,28 @@ on:
   schedule:
     - cron: '0 0 * * MON'
 
-defaults:
-  run:
-    shell: bash
-
 jobs:
   stable:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: ['ubuntu-latest', 'windows-latest']
+        os: ['ubuntu-latest']
         python-version: [3.6, 3.7, 3.8, 3.9, '3.10', 'pypy-3.7']
+        shell: ['bash']
+        include:
+          - os: 'windows-2019'
+            python-version: '3.10'
+            shell: 'bash'
+          - os: 'windows-2019'
+            python-version: '3.10'
+            shell: 'pwsh'
+          - os: 'windows-2019'
+            python-version: '3.10'
+            shell: 'cmd'
 
+    defaults:
+      run:
+        shell: ${{ matrix.shell }}
     steps:
       - uses: actions/checkout@v2
       - name: Set up Python ${{ matrix.python-version }}
@@ -40,11 +50,3 @@ jobs:
         run: tox
         env:
           PLATFORM: ${{ matrix.os }}
-      - name: Test Git (cmd)
-        run: python test/git/test_git.py -v
-        shell: cmd
-        if: ${{ matrix.os == 'windows-latest' }}
-      - name: Test Git (Powershell)
-        run: python test/git/test_git.py -v
-        shell: pwsh
-        if: ${{ matrix.os == 'windows-latest' }}
--- a/src/git/from_vcs.py
+++ b/src/git/from_vcs.py
@@ -22,10 +22,6 @@ def git_pieces_from_vcs(tag_prefix, root, verbose, runner=run_command):
     version string, meaning we're inside a checked out source tree.
     """
     GITS = ["git"]
-    TAG_PREFIX_REGEX = "*"
-    MATCH_ARGS = ["--match", "%s%s" % (tag_prefix, TAG_PREFIX_REGEX)] \
-        if tag_prefix else []
-
     if sys.platform == "win32":
         GITS = ["git.cmd", "git.exe"]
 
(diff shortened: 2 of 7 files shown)


effigies:
Thank you! I will try to review this ASAP, but please feel free to bug me in a day or two if I haven't gotten to it.

.github/workflows/tox.yml
```diff
@@ -19,7 +19,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: ['ubuntu-latest']
+        os: ['ubuntu-latest', 'windows-latest']
         python-version: [3.6, 3.7, 3.8, 3.9, '3.10', 'pypy-3.7']
```
effigies, line 23:
```suggestion
        os: ['ubuntu-latest']
        python-version: [3.7, 3.8, 3.9, '3.10', 'pypy-3.7']
        shell: ['bash']
        include:
          - os: 'windows-latest'
            python-version: '3.10'
            shell: 'bash'
          - os: 'windows-latest'
            python-version: '3.10'
            shell: 'cmd'
          - os: 'windows-latest'
            python-version: '3.10'
            shell: 'pwsh'
```

effigies (commented):
Addressing your question about the explosion in jobs/runtime, this proposes only testing Windows on one Python version, but three shells.

I'm not sure how well tox will work with these other shells, so this might not be the greatest idea. But it feels a bit cleaner than tacking on extra tests for Windows.

effigies:
@mathijsvdv All my suggestions were implemented and debugged in https://github.com/python-versioneer/python-versioneer/tree/maint/windows. If you're good with that, please merge it into your branch.

mathijsvdv:
@effigies  Sounds good to me! I had a quick look at your adjustments and they generally make sense to me. I'll see how the tests are working now and let you know my feedback in the next few days.

src/git/from_vcs.py
```diff
@@ -23,9 +23,11 @@ def git_pieces_from_vcs(tag_prefix, root, verbose, runner=run_command):
     """
     GITS = ["git"]
     TAG_PREFIX_REGEX = "*"
+    MATCH_ARGS = ["--match", "%s%s" % (tag_prefix, TAG_PREFIX_REGEX)] \
+        if tag_prefix else []
```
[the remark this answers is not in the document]
mathijsvdv: We can indeed just put the asterisk directly and define MATCH_ARGS write before the `git describe` call. Your version looks cleaner indeed, thanks

mathijsvdv:
> Addressing your question about the explosion in jobs/runtime, this proposes only testing Windows on one Python version, but three shells.
> 
> I'm not sure how well tox will work with these other shells, so this might not be the greatest idea. But it feels a bit cleaner than tacking on extra tests for Windows.

I agree. One reason I didn't run the whole tox suite was that I was running into some errors when running the whole thing on other shells (and test_git.py contained the tests for the tag prefix). But with your version it seems to be working fine.

Testing fewer Python versions seems a fine compromise to me. One comment I'd make: there was a notable change starting from Python 3.7 that caused the virtual environments to no longer work on Windows until I updated `virtualenv` to `v20` and above. It may be worth testing both Python 3.6 (before) and Python 3.10 (after) for that reason.

.github/workflows/tox.yml
```diff
@@ -12,18 +12,28 @@ on:
...
+        include:
+          - os: 'windows-2019'
+            python-version: '3.10'
+            shell: 'bash'
+          - os: 'windows-2019'
+            python-version: '3.10'
+            shell: 'pwsh'
+          - os: 'windows-2019'
+            python-version: '3.10'
+            shell: 'cmd'
```
mathijsvdv, line 32: Quick question: I don't have a particular preference, but what is the reason for using 'windows-2019' rather than 'windows-latest'?

effigies: 2022 that came out a couple days ago broke building wheels due to not finding the compiler or something. Your tests had run on 2019 and passing, so I reverted for now.

Long run it would be good to figure out what's going on there.

effigies:
Python 3.6 is end of life. I'd suggest we drop that altogether, but we could do 3.7 to test the range.

.github/workflows/tox.yml
mathijsvdv: Right, I remember having similar issues on my local machine. It's C++ Build Tools that you're talking about I presume. I know that's needed to build several packages in Python, so likely this issue will have been flagged in the general Python community already. Let's keep an eye on that.

mathijsvdv:
> Python 3.6 is end of life. I'd suggest we drop that altogether, but we could do 3.7 to test the range.

Fair enough. Since the issue with virtualenvs started with Python 3.7 I think we're pretty covered with Python 3.10 + all the linux tests then, but it can't hurt to cover our bases. I'll leave it to you.

For the rest I have no comments. From my side, we're ready to merge :)

effigies:
Let's do it, and we can tweak later if needed. Thanks a ton for pushing this forward!

hwjohn:
Great work!  Thanks for the fix guys.

mathijsvdv:
No problem, happy to help!

jmartens:
Any chance this will released soon? Current release is broken for some cases when running windows.

effigies:
Yes. I would like to get in #280 before releasing, if anybody has the urge to do a review...

PR #289, opened 2022-03-07
MNT: Drop distutils

### LINKED ISSUE #275: Stop using distutils

effigies:
Distutils is deprecated. The only remaining distutils-dependent code is:

* py2exe command class, waiting on some alternative (see https://github.com/py2exe/py2exe/issues/127)
* build_scripts test (no setuptools replacement; unclear if this is a use case that we should find an alternative approach for)

Closes #275.


jwodder:
@effigies Looks good to me, but I wouldn't take just my word for it.

effigies:
Thanks. I'll give it to the end of the week in case someone else feels inspired...

Issue #188, opened 2018-09-17
should I set `tag_prefix = v`?


whimian:
My git tags starts with "v". Should I set `tag_prefix = v` or just `tag_prefix=`?

PR #294, opened 2022-03-23
ENH: Allow versioneer to be used as a module in PEP-518 mode

### LINKED ISSUE #193: Non-vendored versioneer?
I understand the reason why `versioneer` installs a script in the local directory, but with PEP 517 and PEP 518, we're moving towards a system with explicitly declared build dependencies and running builds in isolated environments.

Currently, if you specify `build-backend="setuptools.build_meta"` in your `pyproject.toml`, you need to also add `sys.path.append(os.dirname(__file__))` into your `setup.py` in order to use `versioneer`.

However, if you're using PEP 517/518, that also means that you don't need to vendor your build dependencies in the source root anymore, you can just declare them in `pyproject.toml`. The difference in semantics between the PEP 517 backend and `python setup.py` commands also gives you a golden opportunity for "opportunistic upgrading". If you provide a version of `versioneer` that you can `pip install`, then anyone using `versioneer` can seamlessly upgrade to PEP 517 by simply adding `versioneer` to their `pyproject.toml`.

Currently, it fails because `pip install versioneer` installs a *different* `versioneer` module than the one that is picked up when you vendor it. I think you can make the PEP 517 transition much easier by giving projects a PyPI package they can depend on (maybe `versioneer` or `versioneer-installed`) that, when installed, puts the `versioneer` people are expecting on their environment's path.

effigies:
As https://github.com/python-versioneer/versioneer-518 proves, the only thing preventing versioneer to be used in a non-vendored fashion and consistent with PEP-518 is making the variables available in the vendored module available via `import versioneer`.

This will not prevent vendored installations but may lead to silent interference. I also haven't added a non-vendored `versioneer setup` command, as I haven't evaluated what all the vendored version does that would be useful in a non-vendored context.

Others' thoughts would be welcome (@bucknerns?).

Closes #193.


effigies:
> The local versioneer.py will take precedence over site packages version so it won't affect any vendored installs. If that is not the case I think having versioneer installed would prevent building vendored versioneer packages since the module doesn't have the get_version call.

I think that is sometimes the case, which leads some projects to add `sys.path.insert(0, ".")` to their `setup.py`. I don't fully understand why it seems to happen to some projects and not others.

yarikoptic:
@jwodder - would you be so kind to provide a review on this PR? My foo in Python build/dist/*-utils and corresponding PEPs is not that strong so can't really assess possible implications.

effigies:
After letting this rest for a bit, I tried using it on a new repository and found this wasn't as ready as I remembered. Got around to adding `versioneer setup` to the base command that skips the `versioneer install` (vendoring) phase. Also added a test that demonstrates that it can work in `python -m build`'s isolated build process. This required a hack to inject the current repository directory as `versioneer @ file:///...`.

Will rebase and update the docs after #188.

effigies:
I think we might want to consider the CLI a bit before pushing this out. Right now the installation process is:

```Shell
versioneer install          # Vendors versioneer then runs setup
python versioneer.py setup  # Rerun setup from vendored versioneer
```

I'm currently proposing that this be possible to replace with `versioneer setup` or `python -m versioneer setup` which will skip the installation step and always use the environment-installed versioneer. This double-use of `setup`, even though it means the same thing, might hide when someone is accidentally invoking the environment versioneer instead of the vendored versioneer, potentially leading to subtle bugs. I think instead we want to revert to `versioneer install` for the main command and `versioneer.py setup` for the vendored module, and add a `--vendor/--no-vendor` flag.

```Shell
versioneer install              # Vendors versioneer then runs setup
versioneer install --vendor     # Same as default
versioneer install --no-vendor  # Non-vendored setup
```

I will push this soon, but adding a comment to explain my reasoning.

effigies:
Took a break to come back to this with fresh eyes. On the whole, I'm happy with it. Would anybody be interested in reviewing and/or test-driving? @bucknerns @yarikoptic?

bucknerns:
I'm checking it out

effigies:
@bucknerns FWIW I currently have to do this to get `python -m build` to work with isolation:

```toml
[build-system]
requires = [
    "setuptools",
    "versioneer @ file:///data/projects/versioneer/",
    "wheel"
]
build-backend = "setuptools.build_meta"
```

Once there's a release, it shouldn't be necessary. And I think `python -m build --no-isolation` should work without the `@`.

effigies:
> I'm checking it out

@bucknerns Any news?

bucknerns:
Looks good, I pushed a copy to test.pypi.org and added to a pyproject.toml and did a `pip install --index-url https://test.pypi.org/simple/ --extra-index-url https://pypi.org/simple/ ./exampleproject`  of an example project to test it out.

bucknerns:
fyi had to version lock it in the pyproject.toml but should be good on official version "versioneer==0.24.dev0"

effigies:
Awesome! Thanks for the review. I'll merge now. I can write some release notes and pushing 0.24 to PyPI today or tomorrow, but feel free to go ahead and do it yourself.

PR #297, opened 2022-04-19
FIX: skip version update on `build_ext` if .py does not exist

### LINKED ISSUE #296: `build_ext` fails if `build_py` was not called before
For example in [aesara](https://github.com/aesara-devs/aesara):

```
$ python setup.py build_ext -j12
running build_ext
UPDATING build/lib/aesara/_version.py
error: [Errno 2] No such file or directory: 'build/lib/aesara/_version.py'
```

mgorny:
Skip version update on `build_ext` command if the respective version
file does not exist.  This could be the case if `build_ext` is called
explicitly without calling `build_py` prior to it.

Closes #296

### DIFF
--- a/src/cmdclass.py
+++ b/src/cmdclass.py
@@ -111,6 +111,11 @@ def get_cmdclass(cmdclass=None):
             # it with an updated value
             target_versionfile = os.path.join(self.build_lib,
                                               cfg.versionfile_build)
+            if not os.path.exists(target_versionfile):
+                print(f"Warning: {target_versionfile} does not exist, skipping "
+                      "version update. This can happen if you are running build_ext "
+                      "without first running build_py.")
+                return
             print("UPDATING %s" % target_versionfile)
             write_to_version_file(target_versionfile, versions)
     cmds["build_ext"] = cmd_build_ext


effigies:
@pkittenis Could I bug you for a review here, since you added the `build_ext` implementation? I'm curious if this is going to cause problems for any prominent use cases.

src/cmdclass.py
```diff
@@ -111,6 +111,10 @@ def run(self):
             # it with an updated value
             target_versionfile = os.path.join(self.build_lib,
                                               cfg.versionfile_build)
+            if not os.path.exists(target_versionfile):
+                print("Warning: %s does not exist, skipping version update" %
+                      target_versionfile)
```
effigies, line 116:
```suggestion
                print(f"Warning: {target_versionfile} does not exist, skipping version update. "
                      "This can happen if you are running build_ext without first running build_py.")
```

effigies (commented):
In the absence of another review, I'll go ahead and merge, but can we make the message more explicit?

mgorny:
Thanks, squashed that change into my commit.

mgorny:
Hmm, I see I need to shorten the lines now. I'll do that in a minute.

mgorny:
No problem. Pushed a fix now.

effigies (approved):

effigies:
Thanks!

PR #305, opened 2022-07-14
Undefined name: VersioneerBadRootError on line 51

cclauss:
`VersioneerBadRootError` is used on line 51 but it is never imported which may result in a NameError being raised instead.

### DIFF
--- a/src/header.py
+++ b/src/header.py
@@ -21,8 +21,7 @@ import sys
 from typing import Callable, Dict
 import functools
 
-from .get_versions import VersioneerBadRootError
-
+class VersioneerBadRootError(Exception): ... # --STRIP DURING BUILD
 
 class VersioneerConfig:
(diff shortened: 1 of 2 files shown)


src/header.py
```diff
@@ -21,6 +21,8 @@
 from typing import Callable, Dict
 import functools
 
+from .get_versions import VersioneerBadRootError
+
```
effigies, line 25:
```suggestion
class VersioneerBadRootError(Exception): ... # --STRIP DURING BUILD
```

effigies (commented):
The way versioneer is set up is actually to concatenate these files into a single file that can be vendored:

https://github.com/python-versioneer/python-versioneer/blob/280fb6ce89af73ac6ced3c66354a16506a57fc6c/setup.py#L51-L74

To avoid multiple imports and definitions, it uses stubs that are then stripped during the build.

PR #308, opened 2022-07-27
FIX: Handle missing tag_prefix gracefully

### LINKED ISSUE #300: Changing tag prefix from a string to null appears to make _version.py use string literal "None"
Using versioneer 0.22

I was trying to fix an issue where I accidentally used tag_prefix in setup.cfg and ended up commenting it out and reinstalling versioneer to the repo:

before:
```
[versioneer]
VCS = git
style = pep440
versionfile_source = myrepo/_version.py
#versionfile_build =
tag_prefix = myrepo
#parentdir_prefix =
```

after:
```
[versioneer]
VCS = git
style = pep440
versionfile_source = myrepo/_version.py
#versionfile_build =
#tag_prefix =
parentdir_prefix = myrepo
```

On reinstall I found this in the _version.py:

```
def get_config():
    """Create, populate and return the VersioneerConfig() object."""
    # these strings are filled in when 'setup.py versioneer' creates
    # _version.py
    cfg = VersioneerConfig()
    cfg.VCS = "git"
    cfg.style = "pep440"
    cfg.tag_prefix = "None"
    cfg.parentdir_prefix = "myrepo"
    cfg.versionfile_source = "myrepo/_version.py"
    cfg.verbose = False
    return cfg
```

Haven't seen this before but it may have to do with changing it to null on an existing installation.

effigies:
This is going to fail CI, but just to put it out there. If someone has time to think clearly about this, be my guest.

Closes #300

### DIFF
--- a/src/header.py
+++ b/src/header.py
@@ -87,7 +87,7 @@ def get_config_from_root(root):
     cfg.versionfile_source = section.get("versionfile_source")
     cfg.versionfile_build = section.get("versionfile_build")
     cfg.tag_prefix = section.get("tag_prefix")
-    if cfg.tag_prefix in ("''", '""'):
+    if cfg.tag_prefix in ("''", '""', None):
         cfg.tag_prefix = ""
     cfg.parentdir_prefix = section.get("parentdir_prefix")
     cfg.verbose = section.get("verbose")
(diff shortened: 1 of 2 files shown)


effigies:
Can't see a downside... Any reviews?

Issue #247, opened 2021-02-28
Options to suppress certain updates


bryevdv:
We always have to manually remove versioneer's edits to these files:
```
	modified:   MANIFEST.in
	modified:   bokeh/__init__.py
```
The change to `MANIFEST.in` is simply unnesessary/superfluous with our configuration. The `__init__.py` is actively interfere. We have additional versioning utilities and hooks on top of `versioneer` to accommodate the extra tests we need to do (we need to simulate different versions at test time) and we handle all that (including getting versioneer's version) in a `util.versions` module, not in `__init__.py` 

It would be nice to have some command line options to `versioneer install` to skip these modifications entirely.

yarikoptic:
ran into the same desire for `__init__.py` since for API preservation purpose we would like to have injected code in `datalad/version.py` for now.  So it felt worthwhile having some `initfile_{source,build}` config option to override.

effigies:
Yup, I run into this as well. What would be the preferred solution? Additional option in `setup.cfg` or a CLI flag like `--skip-snippet`? Something else?

yarikoptic:
Config. I suggested one above for init to follow current config approach.
CLI options would cause inconsistencies across runs.

effigies:
Ah. I can't read.

PR #309, opened 2022-07-27
ENH: Patch versioneer files into manifest at runtime

### LINKED ISSUE #40: MANIFEST.in lines are sometimes (often?  usually?) redundant
in my `setup.py` file I already do this:

```
setup(
    ...
    packages=["mypackage", ...],
)
```

This automatically includes all of the modules in `mypackage`.  It seems likely that most `setup.py` files are written like this or using the `setuptools` equivalent (`find_modules` or something) - though I see how it's not _necessarily_ the case that they are.

`versioneer-install` adds `mypackage/_version.py` to `MANIFEST.in` anyway.  This seems to be harmless but it is redundant and a distraction at best.  At worst it makes things confusing for someone trying to change how the package works (since they must now edit two things in order to change whether/how `_version.py` gets packaged).

It would be pretty cool not to add this file to `MANIFEST.in` if it is redundant to do so.

effigies:
Hack egg_info to modify SOURCES.txt, adding versioneer.py and versionfile_source. versionfile_source is necessary in cases where there's a single module.py file in the root. Admittedly a boutique case, but one we test for.

This removes the need to add these files to `MANIFEST.in`, which has been a source of annoyance (#40, #247, probably others...).

Closes #40.
Addresses part of #247.

### DIFF
--- a/src/cmdclass.py
+++ b/src/cmdclass.py
@@ -177,6 +177,42 @@ def get_cmdclass(cmdclass=None):
                              })
         cmds["py2exe"] = cmd_py2exe
 
+    # sdist farms its file list building out to egg_info, which in turn
+    # farms it out to a subclass of sdist called manifest_maker
+    # To insert versioneer.py into the file listing, we need to create our
+    # own manifest_maker and then patch egg_info to use that
+    if 'egg_info' in cmds:
+        _sdist = cmds['egg_info']
+    else:
+        from setuptools.command.egg_info import egg_info as _egg_info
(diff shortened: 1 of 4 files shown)


effigies:
Thinking about this again, I think using the `manifest_maker` class leaves us more vulnerable to changes in implementation details. By updating `egg_info.filelist` and doing our own write, I think we're less likely to get broken by upstream fixes.

I would appreciate a review, if anybody has the time and interest.

PR #313, opened 2022-08-11
Adequate custom `build_py` command to changes in setuptools v64

abravalheri:
`setuptools` v64 introduces the new PEP 660 hooks, which seems to be incompatible with `versioneer` (as per https://github.com/pypa/setuptools/issues/3501)

The idea of this PR is to try finding a way to make the integration with the new version work properly.

Disclaimer: I am not very familiar with `python-versioneer`, so please forgive me if there is something wrong here.

#### Deep explanation

This means that when `pip install -e .` runs, it will try to use the new `editable_wheel` command instead of `develop`.

Previously, the `develop` command special-cased `build_py` and `build_ext` and treated each command considering the knowledge on what they do.

This was problematic because custom build steps could not run during an editable install.

`editable_wheel` tries to use a more generic approach and fix the problem of custom build steps:
- all the `build` subcommands should run.
- each command have the opportunity to early return by checking if the current build is an editable install or not.

For `versioneer` the implication is that the `build_lib` directory may not contain a `_version.py` file.

### DIFF
--- a/src/cmdclass.py
+++ b/src/cmdclass.py
@@ -69,6 +69,9 @@ def get_cmdclass(cmdclass=None):
     #   then does setup.py bdist_wheel, or sometimes setup.py install
     #  setup.py egg_info -> ?
 
+    # pip install -e . and setuptool/editable_wheel will invoke build_py
+    # but the build_py command is not expected to copy any files.
+
     # we override different "build_py" commands for both environments
     if 'build_py' in cmds:
         _build_py = cmds['build_py']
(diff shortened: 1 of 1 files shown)


effigies:
Thanks @abravalheri! This looks completely sensible. I think our existing guards in `build_ext` are probably already avoiding this case as well. It might be good to try to unify these to be more consistent between commands, but given that this seems to need fixing for many people on a short time scale, I'm inclined to merge and release.

abravalheri:
Thank you very much @effigies. I also added some contingency to `setuptools`. Probably the later releases are much less problematic. Sorry if it caused any disruption.

> It might be good to try to unify these to be more consistent between commands, but given that this seems to need fixing for many people on a short time scale, I'm inclined to merge and release.

In `build_ext` the `inplace` and the `editable_mode` attributes should work in unison with the new editable installs (they are pretty much exchangeable). I don't know if this helps to open the doors for some refactoring.

Issue #265, opened 2021-09-11
Using `branch` in style with no master branch 


martinfleis:
Hi,

just a question - #164 added support for `branch` to be a part of style but it hard-codes `masters` as an expected default branch here https://github.com/python-versioneer/python-versioneer/blob/fe265fce1f6344287b45ee68da62fb0227804c34/src/git/from_vcs.py#L77-L81

Since Github now defaults to `main` branch, there are repos with no `master`. I am curious if this should not be customizable/fetched from git. I cannot report an issue with that or anything, we just found it in the code and we wondering about the potential effects.

effigies:
We should probably make the default branch configurable. We could try automatically setting it with `git config`, so most users don't need to do it manually.

abhishekvalsan-joveo:
Kindly review

PR #316, opened 2022-08-23
ISSUE-265:- make default branch configurable

abhishekvalsan-joveo:
#265

### DIFF
--- a/src/git/from_vcs.py
+++ b/src/git/from_vcs.py
@@ -84,8 +84,13 @@ def git_pieces_from_vcs(tag_prefix, root, verbose, runner=run_command):
 
         # Strip off the leading "* " from the list of branches.
         branches = [branch[2:] for branch in branches]
-        if "master" in branches:
-            branch_name = "master"
+        default_branch, rc = runner(GITS,
+                                ["rev-parse","--abbrev-ref","origin/HEAD"],
+                                cwd=root)
+        # default_branch will be origin/<dafault_branch_name>
+        default_branch = default_branch.split('/')[-1]
+        if default_branch in branches:
+            branch_name = default_branch
         elif not branches:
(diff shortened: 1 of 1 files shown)


abhishekvalsan-joveo:
make default branch configurable

effigies:
Thanks! Thinking about this, I think we need to consider these use cases:

1) New repository with no remote. Probably want the current branch, as you're unlikely to be branching yet.
2) Self-owned repository where you work with just one remote (typically `origin`)
3) Contributing repository where you refer to a canonical fork (generally `upstream` or `origin`, depending on local convention)

I would probably use the following preference order:

1) `git rev-parse --abbrev-ref upstream/HEAD`
2) `git rev-parse --abbrev-ref origin/HEAD`
3) `git rev-parse --abbrev-ref HEAD`
4) `git config init.defaultBranch`
5) `"master"` (git too old to have `init.defaultBranch`)

This value should be computed on installation, not `git_pieces_from_vcs()`, and set in the configuration. We do VCS-specific installation steps in https://github.com/python-versioneer/python-versioneer/blob/master/src/git/install.py.

In `git_pieces_from_vcs()` we should use the configuration variable in place of hard-coded `"master"`.

Now, this is just to get the default branch, which only needs to be done once at installation. When we actually use it, we should use the config variable, since that will be more reliable than each user of the repository naming their remotes the same thing.

https://github.com/python-versioneer/python-versioneer/blob/fa00476f062c65230433d9736a170f575c68f4c9/src/setupfunc.py#L31-L44

We could consider making it `default_branch = {default_branch}`, which would make it easy to detect whether it has been set before or if we should run the inference above.

abhishekvalsan-joveo:
cool, will work on the above comment.

PR #318, opened 2022-08-24
ENH: in verbose mode do not hide stderr of git rev-parse

### LINKED ISSUE #310: pass stderr in verbose mode from running git commands?
e.g. 

```patch
diff --git a/versioneer.py b/versioneer.py
index a48dcce97..da5916f23 100644
--- a/versioneer.py
+++ b/versioneer.py
@@ -1196,7 +1196,7 @@ def git_pieces_from_vcs(tag_prefix, root, verbose, runner=run_command):
     runner = functools.partial(runner, env=env)
 
     _, rc = runner(GITS, ["rev-parse", "--git-dir"], cwd=root,
-                   hide_stderr=True)
+                   hide_stderr=not verbose)
     if rc != 0:
         if verbose:
             print("Directory %s not under git control" % root)
```

since including the fact that I needed `-v` for `pip install` to see this verbose, took awhile to troubleshoot the need to overcome recent git's check for safety:

```shell
$ sudo git "rev-parse" "--git-dir"
fatal: unsafe repository ('/home/travis/build/datalad/datalad' is owned by someone else)
To add an exception for this directory, call:

	git config --global --add safe.directory /home/travis/build/datalad/datalad
The command "sudo git "rev-parse" "--git-dir"" exited with 128.
```
in case of installing via `sudo` from a local git repo (that is how we test datalad on travis).

More of pains could be shared from https://github.com/datalad/datalad/pull/6913

yarikoptic:
Hiding it made it very difficult to figure out why versionner was
failing to figure out the version.

Closes #310

### DIFF
--- a/src/git/from_vcs.py
+++ b/src/git/from_vcs.py
@@ -35,7 +35,7 @@ def git_pieces_from_vcs(tag_prefix, root, verbose, runner=run_command):
     runner = functools.partial(runner, env=env)
 
     _, rc = runner(GITS, ["rev-parse", "--git-dir"], cwd=root,
-                   hide_stderr=True)
+                   hide_stderr=not verbose)
     if rc != 0:
         if verbose:
             print("Directory %s not under git control" % root)


effigies:
Should we add this to all runner() calls in this function?

yarikoptic:
> Should we add this to all runner() calls in this function?

it seems that default is `hide_stderr=False,`

```
$> git grep 'def run_command'
src/git/from_vcs.py:def run_command(): pass  # --STRIP DURING BUILD
src/git/install.py:def run_command(): pass # --STRIP DURING BUILD
src/subprocess_helper.py:def run_command(commands, args, cwd=None, verbose=False, hide_stderr=False,
```
and there is no other overload to `True`  so I do not see any other location to add this

```
$> git grep hide_stderr
src/git/from_vcs.py:                   hide_stderr=not verbose)
src/subprocess_helper.py:def run_command(commands, args, cwd=None, verbose=False, hide_stderr=False,
src/subprocess_helper.py:                                       stderr=(subprocess.PIPE if hide_stderr
test/git/test_git.py:                                 hide_stderr=False, env=None):
```

effigies:
Sounds good.

PR #335, opened 2022-09-13
ENH: Trust the package-installed versionfile

effigies:
Some packages have hacked versioneer to use custom overrides or fallbacks, which makes the process of upgrading somewhat painful, and prevents switching to the non-vendored versioneer. The current versioneer does not execute the contents of `_version.py`, but just greps for certain values before falling back to its own methods.

This PR causes `versioneer.get_versions()` to first attempt to run `_version.get_versions()` before falling back.


effigies:
I'm done trying to update versioneer. If someone wants to pick this up, they can.

PR #337, opened 2022-09-16
Merge `endswith` checks

DimitriPapadopoulos:
[Consider merging `startswith`/`endswith` checks PY-W0077](https://deepsource.io/gh/DimitriPapadopoulos/python-versioneer/issue/PY-W0077/occurrences)

> The methods `startswith` and `endswith` for string instances accept two types of Python constructs as the first argument. The types are:
> - a Python string, or
> - a tuple of strings


effigies:
Thanks!

PR #340, opened 2022-09-23
CI: python-version should be a string, not a float

DimitriPapadopoulos:
This shows clearly with `'3.10'` which is different from `'3.1'`.

### DIFF
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -20,7 +20,7 @@ jobs:
     strategy:
       matrix:
         os: ['ubuntu-latest']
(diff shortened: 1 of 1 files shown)


effigies:
Have you found this causes problems in some situations?

DimitriPapadopoulos:
Yes, when a version ends with `0`.

For documentation purposes, I feel it is better to use explicit strings. Otherwise, other maintainers may for example trigger this error when updating a version from `3.9` to `3.10`.

effigies:
Fair enough.

PR #342, opened 2022-10-15
FIX: mixup between _egg_info and _sdist

DimitriPapadopoulos:
This fixes the following error message that appeared when upgrading Numpy from Versioneer 0.19 to 0.26:
```
  File "setup.py", line 102, in <module>
    cmdclass = versioneer.get_cmdclass(numpy_cmdclass)
  File "/cygdrive/d/a/numpy/numpy/versioneer.py", line 1980, in get_cmdclass
    class cmd_egg_info(_egg_info):
UnboundLocalError: local variable '_egg_info' referenced before assignment
```

### DIFF
--- a/src/cmdclass.py
+++ b/src/cmdclass.py
@@ -189,7 +189,7 @@ def get_cmdclass(cmdclass=None):
 
     # sdist farms its file list building out to egg_info
     if 'egg_info' in cmds:
-        _sdist = cmds['egg_info']
+        _egg_info = cmds['egg_info']
     else:
         from setuptools.command.egg_info import egg_info as _egg_info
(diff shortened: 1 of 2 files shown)


effigies:
Looks correct. I guess we should have a regression test where we prepopulate the cmdclass.

effigies:
@DimitriPapadopoulos Are you interested in adding a test? Would you like guidance?

DimitriPapadopoulos:
Yes, some guidance would be nice. I don't know what the modified code is for. I just fixed the obvious to be able to update Versioneer in another project, without really understanding the purpose of the modified code.

PR #344, opened 2022-10-18
FIX: Always bootstrap in setup.py to avoid incompatibility with old versioneer

### LINKED ISSUE #343: `setup.py` from git fails if `versioneer<=0.23` is installed
Noticed this because Gentoo users hit upgrade problem. If `versioneer==0.23` is installed on the system, `setup.py` run from a git clone/snapshot fails because it tries to use it:

```pytb
$ python setup.py build
  File "/tmp/python-versioneer/setup.py", line 143, in <module>
    VERSION = versioneer.get_version()
              ^^^^^^^^^^^^^^^^^^^^^^
AttributeError: module 'versioneer' has no attribute 'get_version'
```

Having to uninstall the previous version prior to building the new one would be fairly inconvenient, and I can't think of a proper `PYTHONPATH` hack that would prevent the old version from being imported.

mgorny:
Always bootstrap versioneer and use the newly bootstrapped version in `setup.py`.  This avoids compatibility problems when there is an old version of versioneer installed (<= 0.23) that does not define `get_version()` and therefore causes `setup.py` to fail.

Thanks to Chris Markiewicz for the suggested solution.

Fixes #343

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python
 
 import os, base64, tempfile, io
+from importlib import util as ilu
 from pathlib import Path
 from setuptools import setup, Command
 from setuptools.command.build_py import build_py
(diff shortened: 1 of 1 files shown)


effigies:
Looks great, and tests are passing. Thanks very much for your help!

mgorny:
Thanks!

PR #348, opened 2022-10-21
ENH: support built-in tomllib for Python 3.11+

mgorny:
Python 3.11 includes a built-in `tomllib` module (that's compatible with `tomli`).  Use it in new Python versions, and require `tomli` only with older Pythons.


mgorny:
I've also added Python 3.11 to test matrix and regenerated the test C extension using newer Cython, as needed to fix compatibility with Python 3.11.

effigies:
Nice! Thanks very much.

mgorny:
Thanks!

PR #349, opened 2022-11-06
FIX: cx_freeze changed the path to build_exe

spaam:
they changed the path and the name in v6.11

### DIFF
--- a/src/cmdclass.py
+++ b/src/cmdclass.py
@@ -130,7 +130,7 @@ def get_cmdclass(cmdclass=None):
     cmds["build_ext"] = cmd_build_ext
 
     if "cx_Freeze" in sys.modules:  # cx_freeze enabled?
-        from cx_Freeze.dist import build_exe as _build_exe
+        from cx_Freeze.command.build_exe import BuildEXE as _build_exe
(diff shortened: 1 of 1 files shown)


src/cmdclass.py
```diff
@@ -130,7 +130,7 @@ def run(self):
     cmds["build_ext"] = cmd_build_ext
 
     if "cx_Freeze" in sys.modules:  # cx_freeze enabled?
-        from cx_Freeze.dist import build_exe as _build_exe
+        from cx_Freeze.command.build_exe import BuildEXE as _build_exe
```
effigies, line 133: Unless cx_Freeze is very difficult to use any but the latest version of, it probably makes sense to support a range of versions.

```suggestion
        try:
            from cx_Freeze.command.build_exe import BuildEXE as _build_exe
        except ImportError:  # cx_Freeze < SOME_VERSION
            from cx_Freeze.dist import build_exe as _build_exe
```

Callek: Created a new PR with this suggestion.  Thanks @spaam for the initial PR.

Callek:
This was fixed in #386

PR #352, opened 2022-12-16
Apply pyupgrade and refurb suggestions

DimitriPapadopoulos:


src/header.py
```diff
@@ -1,4 +1,3 @@
-
```
effigies, line 1: This whitespace is significant.

src/render.py
```diff
@@ -1,4 +1,3 @@
-
```
effigies, line 1: And here

src/setupfunc.py
```diff
@@ -1,4 +1,3 @@
-
```
effigies, line 1: And here

src/header.py
DimitriPapadopoulos: That's what I suspected, but it's not clear why. Is additional code added in front of it?

effigies: Yes, these files are compiled into the single-file `versioneer.py` file.

DimitriPapadopoulos: The alternative would be to add the extra line when compiling the `versioneer.py` file in `setup.py`. But then, why change something that works?

DimitriPapadopoulos: Fixed, but I still need to look into the files under `test`.

src/from_file.py
```diff
@@ -43,7 +43,7 @@ def write_to_version_file(filename, versions):
     contents = json.dumps(versions, sort_keys=True,
                           indent=1, separators=(",", ": "))
     with open(filename, "w") as f:
-        f.write(SHORT_VERSION_PY % contents)
+        f.write(SHORT_VERSION_PY.format(contents))
 
-    print("set %s to '%s'" % (filename, versions["version"]))
+    print("set {} to '{}'".format(filename, versions["version"]))
```
effigies, line 48: Might as well use an f-string if we're here.

```suggestion
    print(f"set {filename} to {versions['version']:r}")
```

src/from_parentdir.py
```diff
@@ -19,8 +19,8 @@ def versions_from_parentdir(parentdir_prefix, root, verbose):
         root = os.path.dirname(root)  # up a level
 
     if verbose:
-        print("Tried directories %s but none started with prefix %s" %
-              (str(rootdirs), parentdir_prefix))
+        print(f"Tried directories {str(rootdirs)} "
+              f"but none started with prefix {parentdir_prefix}")
```
effigies, line 23: `str()` is called implicitly.

```suggestion
        print(f"Tried directories {rootdirs} "
              f"but none started with prefix {parentdir_prefix}")
```

src/git/from_vcs.py
```diff
@@ -111,18 +111,16 @@ def git_pieces_from_vcs(tag_prefix, root, verbose, runner=run_command):
...
             # unparsable. Maybe git-describe is misbehaving?
-            pieces["error"] = ("unable to parse git-describe output: '%s'"
-                               % describe_out)
+            pieces["error"] = f"unable to parse git-describe output: '{describe_out}'"
             return pieces
 
         # tag
         full_tag = mo.group(1)
         if not full_tag.startswith(tag_prefix):
+            error = f"tag '{full_tag}' doesn't start with prefix '{tag_prefix}'"
```
effigies, line 120: No clear need for a temporary variable just to print and then assign...

```suggestion
            pieces["error"] = f"tag '{full_tag}' doesn't start with prefix '{tag_prefix}'"
```

src/git/from_vcs.py
```diff
@@ -111,18 +111,16 @@ def git_pieces_from_vcs(tag_prefix, root, verbose, runner=run_command):
...
             if verbose:
-                fmt = "tag '%s' doesn't start with prefix '%s'"
-                print(fmt % (full_tag, tag_prefix))
-            pieces["error"] = ("tag '%s' doesn't start with prefix '%s'"
-                               % (full_tag, tag_prefix))
+                print(error)
+            pieces["error"] = error
```
effigies, line 123:
```suggestion
                print(pieces["error"])
```

effigies (approved):
All seems fine. Some small suggestions, but I'm okay to merge as-is.

src/from_file.py
DimitriPapadopoulos: But then we're modifying functionality, by changing `'%s'` to `{...}` instead of `'{...}'`, aren't we?

src/git/from_vcs.py
DimitriPapadopoulos: Indeed, this was for consistency with later fix.

src/from_file.py
DimitriPapadopoulos: Also, I suspect you meant `!r` instead of `:r`. It will call [`repr()`](https://docs.python.org/3/library/functions.html#repr), but then how is that helpful?
```python
>>> i = 2
>>> 
>>> f"We print 'i' as '{i}'"
"We print 'i' as '2'"
>>> 
>>> f"We print 'i' as {i!r}"
"We print 'i' as 2"
>>> 
```
We might use triple quotes instead:
```python
f"""set {filename} to '{versions["version"]}'"""
```
But then it gets unreadable.

DimitriPapadopoulos:
I still need to fix these errors, don't I?
```
  git_version.py:335:89: E501 line too long (90 > 88 characters)
  versioneer.py:809:89: E501 line too long (90 > 88 characters)
  versioneer.py:1330:89: E501 line too long (90 > 88 characters)
```

PR #353, opened 2022-12-17
MNT: Overwrite file instead of deleting/writing it

DimitriPapadopoulos:
Why create a new inode? Overwrite the contents of the existing inode.

### DIFF
--- a/src/from_file.py
+++ b/src/from_file.py
@@ -15,7 +15,6 @@ def get_versions():
     return json.loads(version_json)
 """
 
-import os # --STRIP DURING BUILD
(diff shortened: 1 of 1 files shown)


effigies:
Was curious why it was this way. Seems like it was added on the second commit (https://github.com/python-versioneer/python-versioneer/commit/2d60f553902c308475070ff57eaad19d4f19ba88).

The main thing I can think of would be that if there's a chance the file was linked/symlinked, you could end up with a file modified at a distance. Since this is happening in the build directory, a copy is most likely, so that danger seems low.

PR #354, opened 2022-12-17
MNT: Remove pylint hint

DimitriPapadopoulos:
It should be possible to fix some issues reported by pylint instead of disregarding them.


effigies:
Sure. If we're going to manage this list, though, we should go ahead and add pylint to `tox.ini`.

Issue #211, opened 2020-08-03
Help needed?


yarikoptic:
Versioneer got loved by many, thus of all **big thank you** @warner for it!

But is it the right assessment of the situation that its maintenance/development is somewhat "froze up"?

I know e.g. that some projects started to use custom forks (https://github.com/nipy/nibabel/blob/master/versioneer.py) and thus duplicating efforts etc. 

May be you @warner  would accept co-maintainership and/or even move it from under your personal wing under a dedicated https://github.com/python-versioneer organization ?  Not that it miraculously would resolve all the issues at once, but I think it might pave the way forward.

Thank you in advance for reply.

Ping @effigies @jwodder @mih

bashtage:
While this is clearly the best path, the alternative would be to setup an org with an unofficial but hopefully widely adopted fork that could become a defacto successor.

yarikoptic:
Yes @bashtage , that would be they path to proceed if no reply follows.

bucknerns:
I agree an org where multiple participants pave the way forward would be better.
I decided to jump on this and created the org.

bashtage:
> created the org.

link to repo?

bucknerns:
I invited @yarikoptic and @bashtage as owners of the org. https://github.com/python-versioneer

effigies:
Care to open up issues on that branch? IMO, a good first step would be to curate some of the issues from here.

In the meantime, I would say the most important questions are:

1) What versions of Python should the fork support?
2) Continue with vendoring, disable it, or make optional?

Also worth figuring out an overall scope to the fork, to help determine whether a request should be acted on. I think the legitimacy of the fork will be helped by a maintenance team that's on the same page.

Speaking of which, I'm happy to help with this effort, as well.

bucknerns:
So currently we don't have access to the pypi project so pushing new versions of versioneer is not an option. I think we should split apart the versioneer.py file in the versioneer-518 repo remove python2 compatibility and support 3.6+. Also I would remove the vendoring and just use the pep 518 style.

effigies:
IMO including PEP numbers is needless cognitive overhead. For users not really deep in the packaging weeds (which I don't think we should assume people wanting to use versioneer are), 518 isn't going to be very meaningful and invites versioneer-517 as a typosquat. Would you be open to forking to a new name?

A couple that come to mind that are not yet claimed on PyPI are:

* Postfixed
  * versionaut (a la astro-/cosmo-naut)
[...]

PR #361, opened 2023-01-24
FIX: add message info when loading toml failed

GCS-ZHN:
Current version of versioneer will automatically skip it if load `pyproject.toml` failed, and try to load config from `setup.cfg`. 
If someone use versioneer in his project and unluckily there are some mistakes on his `pyproject.toml`, he will get a `FileNotFoundError` about the miss of `setup.cfg`. This is confusive to him.

So I add a info message to make him realize his `pyproject.toml` contains some mistakes(e.g. key depulication). Otherwise, he may spend a lot of time to find why versioneer try to load `setup.cfg` file which not used in his project.

### DIFF
--- a/src/header.py
+++ b/src/header.py
@@ -91,8 +91,9 @@ def get_config_from_root(root):
             with open(pyproject_toml, 'rb') as fobj:
                 pp = tomllib.load(fobj)
             section = pp['tool']['versioneer']
-        except (tomllib.TOMLDecodeError, KeyError):
-            pass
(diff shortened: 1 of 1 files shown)


effigies:
I merged this, because I see nobody's responded in a month and a half. It seems with me stepping down, this project is effectively unmaintained again. I might bug people on https://github.com/python-versioneer/python-versioneer/issues/211 to see what's needed to get it released.

Issue #298, opened 2022-04-20
Add type annotations


dorak88783:
The current code doesn't have type annotations. It would be an improvement if they were added.

effigies:
Please feel free to submit a PR!

PR #365, opened 2023-04-16
Add Type Annotations for `git_version.py`

Callek:
Over in #298 it was suggested we wanted type annotations and that a PR would be welcome.

This PR does roughly the minimum to enable type annotations on `git_version.py`

If I have further time I'll create new PRs for the rest of the src repo, add typing in tests, as well as do some better Typing in a few places (TypedDict, better typing on the decorators and vcs hooks, etc.)

But in the meantime wanted to submit this.

Additional changes that are not strictly typing:
* Use `exception OSError as e` instead of relying on the `sys.exc_info` to get the exception class.  This is ok now that py2.7 is no longer supported [and necessary for better typing since the former method typed it as `BaseException` which didn't have `errno` as a property.
* Supported my local testing on mac (using the linux tox target) by replacing `.` with `_` in the binary wheel name expectation.
* Converted the class/function definitions in the various files (where `# --STRIP DURING BUILD` was used) to actual imports that didn't fail the tests.  This so that python editors (such as VSCode) gave better type hinting in the various files, without the need to type-hint all the redefinitions themselves

### DIFF
--- a/src/cmdclass.py
+++ b/src/cmdclass.py
@@ -1,9 +1,7 @@
 import os, sys # --STRIP DURING BUILD
-LONG_VERSION_PY = {} # --STRIP DURING BUILD
-def get_versions(): pass # --STRIP DURING BUILD
-def get_root(): pass # --STRIP DURING BUILD
-def get_config_from_root(): pass # --STRIP DURING BUILD
-def write_to_version_file(): pass # --STRIP DURING BUILD
+from .header import LONG_VERSION_PY, get_root, get_config_from_root # --STRIP DURING BUILD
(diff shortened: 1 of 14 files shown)


CallekTest:
> Overall LGTM. I would consider adding type checking commands to the tox script, or possibly adding a type-checking environment.

We do have it at least for the `git_version.py` output. I'll need to add a check for the `versioneer.py` output in a followup PR. https://github.com/python-versioneer/python-versioneer/blob/9d14bb5175847cd1116eac3a0e3ef80ae9713856/tox.ini#L59

Callek:
(oops commented with a test account I was using for something else for work)

PR #371, opened 2023-06-18
Detect pyproject.toml as project root

gamecss:
Sometimes I use some package build tools which doesn't require setup.py, for example, pdm.backend. It doesn't need a setup.py, instead, it'll build wheel directly.
So with this pr, I can do this:
```toml
[tool.pdm.version]
source = "call"
getter = "versioneer:get_version"
```
Then when I run `pdm build`, pdm will read version from 'versioneer.get_version()'.
However, this doesn't work now because versioneer requires setup.py or versioneer.py to detect project root. It cannot find project root without setup.py or versioneer.py. When I using build-time dependency mode and "setup.py-less" build backend, either setup.py or versioneer.py doesn't exist, but the project root is still here! So this pull request made versioneer able to detect project root by pyproject.toml.

### DIFF
--- a/src/header.py
+++ b/src/header.py
@@ -54,13 +54,23 @@ def get_root() -> str:
     """
     root = os.path.realpath(os.path.abspath(os.getcwd()))
     setup_py = os.path.join(root, "setup.py")
+    pyproject_toml = os.path.join(root, "pyproject.toml")
     versioneer_py = os.path.join(root, "versioneer.py")
-    if not (os.path.exists(setup_py) or os.path.exists(versioneer_py)):
+    if not (
+        os.path.exists(setup_py)
+        or os.path.exists(pyproject_toml)
+        or os.path.exists(versioneer_py)
+    ):
         # allow 'python path/to/setup.py COMMAND'
         root = os.path.dirname(os.path.realpath(os.path.abspath(sys.argv[0])))
         setup_py = os.path.join(root, "setup.py")
+        pyproject_toml = os.path.join(root, "pyproject.toml")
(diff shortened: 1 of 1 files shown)


effigies:
Huh, I had no idea it was possible to use this outside setuptools. Would you mind contributing docs showing how to do this?

@Callek Any interest in shepherding this one in?

gamecss:
Only PDM (build-backend: pdm.backend) can read version by call function now (I'm not sure if hatchling works because it seem use regex to get the version). Config for pdm.backend included in the first comment.

src/header.py
```diff
@@ -54,13 +54,23 @@ def get_root() -> str:
...
-    if not (os.path.exists(setup_py) or os.path.exists(versioneer_py)):
+    if not (
+        os.path.exists(setup_py)
+        or os.path.exists(pyproject_toml)
+        or os.path.exists(versioneer_py)
+    ):
         # allow 'python path/to/setup.py COMMAND'
         root = os.path.dirname(os.path.realpath(os.path.abspath(sys.argv[0])))
         setup_py = os.path.join(root, "setup.py")
+        pyproject_toml = os.path.join(root, "pyproject.toml")
```
Callek, line 67: I don't necessarily think doing the path translation is necessary here, but happy to keep it for consistency.  (I actually suspect we should rewrite some of this logic a bit, since it *defaults* to cwd() rather than the actual target of the package.)

Callek (approved):
I apologize for the delay here - had a bunch of things going on that made this take me longer than I wanted to.

But I'm happy to have this merge in.  Thank you for the patch!

PR #376, opened 2023-08-20
Add support for python 3.12

### LINKED ISSUE #375: Doesn't work with Python 3.12
I'm trying to build [BuildStream](https://github.com/apache/buildstream/) on Fedora 39 which has Python 3.12, and [it fails](https://github.com/apache/buildstream/issues/1855) :disappointed: 

Once again, fedora is first on the latest and greatest, and we're the one finding issues in a lot of software. :shrug:

I think that issue in BuildStream should give you all you need to fix the issue in versioneer:

```
commit 06ca13b4181c9140a536a72086a324e242ec249b (HEAD -> main)
Author: Mathieu Bridon <mathieu@hashbang.coop>
Date:   Tue Aug 8 09:28:01 2023 +0200

    Try to fix #1855

diff --git a/versioneer.py b/versioneer.py
index 1c97e02..5751584 100644
--- a/versioneer.py
+++ b/versioneer.py
@@ -340,9 +340,16 @@ def get_config_from_root(root):
     # configparser.NoOptionError (if it lacks "VCS="). See the docstring at
     # the top of versioneer.py for instructions on writing your setup.cfg .
     setup_cfg = os.path.join(root, "setup.cfg")
-    parser = configparser.SafeConfigParser()
-    with open(setup_cfg, "r") as f:
-        parser.readfp(f)
+    if sys.version_info.major == 3 and sys.version_info.minor >= 12:
+        parser = configparser.ConfigParser()
+        with open(setup_cfg, "r") as f:
+            parser.read_file(f)
+
+    else:
+        parser = configparser.SafeConfigParser()
+        with open(setup_cfg, "r") as f:
+            parser.readfp(f)
+
     VCS = parser.get("versioneer", "VCS")  # mandatory
 
     def get(parser, name):
```

Can this be merged soon ? :zany_face:

Callek:
~This should fix #375~  (375 was an already fixed issue, but looks like there may still be more necessary for 3.12)


Callek:
Looks like part of the issues is due to one of the tests that uses `Cython` and has a `.c` that was generated with `Cython` `0.29.32` and there was a Py3.12 fix in a newer version.  I'm going to have to solve that for my own confidence before I release a (official) Py3.12 supported `versioneer` version.

Callek:
Looks like we're passing CI for 3.12 now, there are still some depraction warnings and such that should probably be taken care of before we publish a `0.30` but this updates the main repo to ensure we don't regress 3.12 support at least.

test/git/test_invocations.py
```diff
@@ -58,6 +58,12 @@ def make_venv(self, mode):
             self.run_in_venv(venv_dir, venv_dir,
                              'pip', 'install', '-U',
                              'pip', 'wheel', 'packaging')
+        if sys.version_info.major == 3 and sys.version_info.minor >= 12:
+            # Py 3.12 no longer default-installs setuptools in environments
+            # See PEP-632
+            self.run_in_venv(venv_dir, venv_dir,
+                             'pip', 'install', '-U',
+                             'setuptools')
```
effigies, line 66: I'd just go ahead and merge this into the pip install above. I don't think it's important to test that you can get away without installing setuptools on other versions.

effigies (approved):
All makes sense to me.

test/git/test_invocations.py
Callek: My thinking was "Its probably important enough to make sure that the default-installed setuptools on the other versions works"  such that this installs "latest" setuptools from pypi.

effigies: Your call. That's fine by me.

PR #378, opened 2023-09-07
CI: End testing on the end-of-life Python 3.7

### LINKED ISSUE #377: CI broken due to downgraded flake8
Don't have time to investigate tonight but the weekly ci run just failed across multiple python versions. 

CI seems to be suddenly installing an ancient flake8 version [older than it did last week]

effigies:
Looking at #377, it was hard to tell whether 3.7 was the only affected runner. It may be some effect of running tests on an EOL Python, or there could be something deeper.

If it's just 3.7 failing, I would suggest just dropping it.

Closes #377.

### DIFF
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -21,7 +21,7 @@ jobs:
       fail-fast: false
       matrix:
         os: ['ubuntu-latest']
-        python-version: ['3.7', '3.8', '3.9', '3.10', '3.11', '3.12', 'pypy3.9']
+        python-version: ['3.8', '3.9', '3.10', '3.11', '3.12', 'pypy3.9']
         shell: ['bash']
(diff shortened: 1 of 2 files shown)


effigies:
@Callek What do you say to just dropping 3.7 from the test matrix?

Callek:
> @Callek What do you say to just dropping 3.7 from the test matrix?

Probably worth doing given it's ci failure. I was hoping to formally announce some sort of support plan that did map to python support, since we didn't have that. 

But yea, imo, we can drop 3.7 from the test matrix and note that formal support has ended, though we haven't explicitly made changes to drop 3.7 yet.

Callek (approved):
Thanks for taking the extra time!

effigies:
You can tell I really should be doing something else...

PR #386, opened 2023-12-31
Fix cx_Freeze >= 6.11

Callek:
This builds upon (and fixes) PR #349

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,7 @@ This release adds formal support for Python 3.12, and PyPy 3.10.
 Also removes supports for Python 3.7.
 
 * BREAKING CHANGE: Drop formal support for Python 3.7. by @Callek (#385)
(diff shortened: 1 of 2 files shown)


effigies (approved):
Lgtm. Is it plausible to add a test? I know nothing about cxfreeze, so no worries if not. It's survived long enough without one.

Callek:
> Lgtm. Is it plausible to add a test? I know nothing about cxfreeze, so no worries if not. It's survived long enough without one.

Not super easy, but i am (slowly) working on adding tests and doing some research around what we do support, to have clear answers.   Especially with py3.12 completely deprecating setup.py.

I might cut the release before I have that though.

effigies:
To be clear, setup.py is not deprecated, only using it as a command is. It is still the way to configure setuptools when a purely declarative configuration is insufficient.

PR #388, opened 2024-01-17
Print warnings/informative messages to stderr not stdout to avoid interferring with setuptools etc

yarikoptic:
In datalad-container we use

    python setup.py --name

to programmatically determine main package name in the github action. Unfortunately recent upgrade of versioneer made it not usable due to issuing its mere informative messages to stdout which we expect to contain only target value

    ❯ python setup.py --name 2>/dev/null
    Failed to load config from /home/yoh/proj/datalad/datalad-container/pyproject.toml: 'versioneer'
    Try to load it from setup.cfg
    datalad_container

    ❯ echo $?
    0

I think it is more logical and consistent with other tools (e.g. git-annex, git etc) to output similar diagnostic logging messages to stderr.

### DIFF
--- a/src/header.py
+++ b/src/header.py
@@ -89,7 +89,8 @@ def get_root() -> str:
         vsr_dir = os.path.normcase(os.path.splitext(versioneer_py)[0])
         if me_dir != vsr_dir and "VERSIONEER_PEP518" not in globals():
             print("Warning: build in %s is using versioneer.py from %s"
-                  % (os.path.dirname(my_path), versioneer_py))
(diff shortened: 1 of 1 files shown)


Callek:
This sounds logical to me, I'll plan to review this week. 

@effigies - just to check, do you recall any specific reason we hadn't used stderr for these before?

effigies:
No specific reason. This seems fine to me.

Callek (approved):
I noticed there are a bunch more `print`'s that probably should print to stderr, but this is a good start.

Thank You!

PR #394, opened 2024-08-23
Use gztar for sbuild, fixes #393

sudhackar:
As pip maintainers mentioned `tar` is a legacy format and is not supported by 24.0 - we need this to pass the tests

PR #395, opened 2024-09-01
Further fix tests by using setuptools copy of 'bdist_wheel' functions, if available

Callek:


Callek:
Noticed this when reviewing #394

Callek:
While I can merge this without review - I'm a big fan of reviews on code changes, even my own.

@mwtoews (you emoji reacted here) and @sudhackar (most recent person to submit a PR) -  Do either of you want to review this in @effigies 's current absense?

mwtoews (approved):
Using a recent version of `wheel`, the original import raises an exception:
```pycon
>>> import wheel
>>> wheel.__version__
'0.44.0'
>>> from wheel.bdist_wheel import get_abi_tag, get_platform
  File "<stdin>", line 1, in <module>
ImportError: cannot import name 'get_abi_tag' from 'wheel.bdist_wheel' (/path/to/wheel/bdist_wheel.py)
```
This PR resolves this ImportError in a forward-compatible way.

---

My only nit is that the comment has mixed single `` ` ``  and double ` `` ` backticks. Maybe just use unquoted setuptools and wheel in the comments instead.

effigies (approved):
Looks reasonable, thanks for your patience.

sudhackar:
LGTM too!

PR #402, opened 2024-12-26
feat: hatchling backend

kiyoon:
Supports hatchling backend.

I think this should work but the only problem is that I couldn't get the setup.py to work..

Currently it seems that the only file exists in the build is `versioneer.py` but we need

```bash
📂 versioneer/
│  🐍 __init__.py    # originally versioneer.py
└  🐍 hatchling.py
```


kiyoon:
But I really do think that the current setup procedure is confusing and needs to be addressed for further contributors to work on this project.

Issue #390, opened 2024-01-17
`_version.py` uses variable 'dispcmd' before assignment


dougiesquire:
`versioneer install --no-vendor` produced a `_version.py` file that uses the variable `dispcmd` before assignment. 

My project is [here](https://github.com/dougiesquire/payu/tree/dc445a170c4c0316e20fd1cc4d8e402dfc6dc19d) and the offending line in the `_version.py` file is [here](https://github.com/dougiesquire/payu/blob/dc445a170c4c0316e20fd1cc4d8e402dfc6dc19d/payu/_version.py#L112). Should [L101](https://github.com/dougiesquire/payu/blob/dc445a170c4c0316e20fd1cc4d8e402dfc6dc19d/payu/_version.py#L101) be outside of the `try-catch` block?

effigies:
Yes, it should.

PR #403, opened 2025-01-20
Fix used-before-assignment bug in `_version.py`

dougiesquire:
Simple fix for #390

### DIFF
--- a/_version.py
+++ b/_version.py
@@ -97,8 +97,8 @@ def run_command(
         popen_kwargs["startupinfo"] = startupinfo
 
     for command in commands:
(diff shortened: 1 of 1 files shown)


Callek (approved):
from what I can tell, `dispcmd` is only ever used when its been successfully defined, however I can see why some linters *might* have issues here...

I don't see a real problem moving it outside of the try/except though.

Callek:
The 3.8 error doesn't matter because we just dropped 3.8 support, and the 3.12 errors are mypy-only errors right now that I'm looking into.

Thanks for the patch

PR #405, opened 2025-02-09
Mypy Setuptools fix

Callek:
The update to the latest `types-setuptools` breaks because it defines `user_options` as a `ClassVar`... fixed our typing.

This also has a ride-along that I missed when reviewing the "drop py3.8" logic


effigies (approved):
Lgtm.

PR #410, opened 2025-08-21
Use consistent hash length

### LINKED ISSUE #156: fix number of digits used in abbreviated git commit hashes
Git now uses a dynamic length for the number of digits in hashes printed by ``git describe``:
https://github.com/git/git/commit/e6c587c733b4634030b353f4024794b08bc86892

For the purposes of versioneer, where the commit remains unambiguous due to the tag and distance being included in the output, I propose fixing the number of digits to some value (we use 8 in our project). This can be achieved by adding ``--abbrev=8`` to the git command line (https://github.com/m-labs/artiq/commit/076e4119f744c07afaa6f6cd840a45dfcd13718e).

The default git behavior can result in annoying inconsistencies and bugs. In our project, we have several conda packages built from the same repository. For reasons that I do not fully understand, git ended up using different number of characters for the hash lengths in the version numbers of each, causing further problems later on.

dan-blanchard:
Fixes #156 by always using `--abbrev=7` in `git describe` calls to ensure hashes are consistent.

I went with 7 characters, because that's also the length used when using `git rev-parse --short HEAD` when `describe` is not available.

This will make sure that the version strings generated by `versioneer` are consistent across different machines and environments.

(I recently ran into an issue with a development version of `dask` where the short hash varied between environments)

### DIFF
--- a/_version.py
+++ b/_version.py
@@ -283,7 +283,7 @@ def git_pieces_from_vcs(
     # if there is a tag matching tag_prefix, this yields TAG-NUM-gHEX[-dirty]
     # if there isn't one, this yields HEX[-dirty] (no NUM)
     describe_out, rc = runner(GITS, [
-        "describe", "--tags", "--dirty", "--always", "--long",
+        "describe", "--tags", "--dirty", "--always", "--long", "--abbrev=7",
         "--match", f"{tag_prefix}[[:digit:]]*"
     ], cwd=root)
     # --long was added in git-1.5.5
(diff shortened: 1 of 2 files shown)


Callek (approved):
Looks good overall. This does fix the tagged-path inconsistency caused by git describe’s dynamic default abbreviation length. One nit: the PR text should not say this “ensures” fixed-length hashes, because --abbrev=7 still allows Git to emit more characters when needed for uniqueness. Also, the fallback rationale should reference rev-parse HEAD + [:7], not rev-parse --short HEAD. I verified git describe --abbrev=<n> is present at least in Git 2.0.5; I couldn’t confirm the exact first release from authoritative docs

I'll likely merge this later this week (or next) and cut a release.

Thanks for the patch, and sorry for the delay!

PR #417, opened 2026-05-05
Remove incorrect Py_DECREF for marker

wr-web:
https://github.com/python-versioneer/python-versioneer/issues/416


wr-web:
@Callek This should be better.

PR #420, opened 2026-05-26
Commit some pins to fix tests as they are

Callek:
This just unbreaks tests for now.  I'm working on an additional PR that doesn't require the pin


effigies (approved):
Looks good. One long term approach could be to create a legacy selector for tox that pins old setuptools, and then mark tests as legacy, which only run if a compatible setuptools is installed.

